Issue #19, opened 2019-09-28
movzx   ecx, al wrong disassembled


liiir1985:
0F B6 C8 is wrongly decoded as:
movzx eax, byte [eax]

but actually it should be:
movzx eax, al

Washi1337:
Hi,

Thanks for the report. Unfortunately, at the moment, the x86 encoder/decoder namespace has very low priority in terms of further development, especially since I started working on AsmResolver v4 (check out the v4 branch for progress).

For the time being, you might be interested in using [Iced](https://github.com/0xd4d/iced/) from 0xd4d instead. It is tested a lot more and supports more features of the x86 instruction set.

Thanks for understanding.

liiir1985:
Thanks for the reply, but unfortunately Iced doesn't suit my requirement, because I also need the ability to load PE files, this is the main reason why I've chosen AsmResolver

Do you have any roadmap regarding the v4 version? what's the difference?

I think I can look into the decoding problem once I got time, and I'll make a PR if I managed to fix the problem

Washi1337:
Version 4 is a complete rewrite of the library, with a much easier and more consistent API. All of the documentation will live in the `docs` folder, or view  https://asmresolver.readthedocs.io/ for an online, rendered version of it. It is still early in development and names of classes and structures are subject to change.

Some good news is that it will still support loading native binaries, and will still expose low level structures as well as high level structures, however I have not yet decided exactly on how to do the native x86 instruction encoding/decoding, if it will be in there at all. Chances are that I will be offloading that to another library, providing the user with the ability to swap out encoders/decoders for different machine languages.

Until v4 is ready to be merged with master, I will still accept PRs for v3, especially since some projects still use it. If you can find the bug in the decoder, I'd be happy to merge your PR for it. The code might actually end up in the newer version as well.

PR #20, opened 2019-09-29
Fixed the decoding problem of some multibyte opcodes #19

liiir1985:
Fixed the problem of decoding movzx ecx, al
It seems the multibytes opcodes are all missing registertoken


Washi1337:
Thanks for taking the time to spot the bug! Will merge and close the issue ;)

PR #41, opened 2020-03-10
Extension method to check for compiler generated members

766F6964:
Provides an extension method to check if a member was automatically generated by the compiler by checking if the  _CompilerGeneratedAttribute_ CustomAttribute is set.


Washi1337:
Great thanks!

PR #75, opened 2020-06-29
AssemblyDefinition is now able to write the ManifestModule to a stream. ModuleDefinition now has .ToPEImage()

Charterino:
Added 3 Write() methods to allow AssemblyDefinition to write the manifest module to a stream.


Charterino:
ModuleDefinition now has 2 new methods:
.ToPEImage() which builds a PEImage using ManagedPEBuilder by default 
.PEImage(IPEImageBuilder) which builds a PEImage using the specified builder.

PR #87, opened 2020-07-08
Bugfix/methodsig as operand

TobitoFatitoRE:
Asmresolver seemed to not check if operand is methodsig while calculating push&pops.


Washi1337:
It never checks for a `MethodSignature` operand because the default operand resolver never assigns `CilInstruction.Operand` to a `MethodSignature`. In the case of the `calli` opcode, the operand is of type `StandAloneSignature`, which contains the referenced `MethodSignature`.

See:
- https://asmresolver.readthedocs.io/en/latest/dotnet/managed-method-bodies.html#basic-structure-of-cil-instructions
- ECMA-335, Partition I,  section II.22.36, page 243

Issue #95, opened 2020-09-19
Incorrect Section FileOffset Calculation


SapientGuardian:
Hello! First I'd like to thank you for creating this library and others; they're more or less exactly what I need for a project I'm working on, and there is a serious lack of these sorts of projects written in C#.

I have a simple test in which I'm attempting to round-trip the attached EXE  (not malware; safe to execute and I can provide source if you want it), which is a very simple C++ program compiled with Visual Studio 2019 and not designed to be "tricky" in any way, through AsmResolver and verify that the resulting binary is identical:

```
            var originalBytes = Properties.Resources.NativeMemoryDemos;
            var peFile = PEFile.FromBytes(originalBytes);

            var msOutput = new MemoryStream();
            var output = new PEFile(peFile.DosHeader, peFile.FileHeader, peFile.OptionalHeader);
            foreach (var section in peFile.Sections)
            {
                var newSection = new PESection(section);
                output.Sections.Add(newSection);
            }
            output.Write(new BinaryStreamWriter(msOutput));
            Assert.Equal(originalBytes, msOutput.ToArray());
```

The output is identical, except for the FileOffset of the first section (`.textbss`), which has changed from 0 to 0x400, which is the same offset as the second section (`.text`) as illustrated in the screenshot from FileAlyzer below:

![image](https://user-images.githubusercontent.com/14796343/93686172-7ee74680-fa82-11ea-9631-44e1ae0fe18e.png)

I'm a novice when it comes to the PE format, so I apologize if I've incorrectly analyzed the problem, but I believe it to be the result of `AlignSections()` not taking into account the rule that "When a section contains only uninitialized data, this field should be zero." (https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers). 

In `AlignSections()` it does this:
```
  var section = Sections[i];

                uint fileOffset = i > 0
                    ? Sections[i - 1].FileOffset + Sections[i - 1].GetPhysicalSize()
                    : OptionalHeader.SizeOfHeaders;
```

which forces the offset of the first section to 0x400, as that is the value of `OptionalHeader.SizeOfHeaders`.

I think the correct behavior would be to check if the section has the `ContentUninitializedData` flag set and force fileOffset to 0, regardless of whether it is the first section or not.

This brings up an interesting problem though, which is that in both the current and proposed behavior, I (the user) don't have a chance to write an arbitrary value into the table if I so desire.

[...]

Issue #64, opened 2020-06-04
Support for reading PE files in mapped mode


Washi1337:
**Summary:**
Currently `PEFile` always assumes an unmapped mode of the file, where RVAs need to be translated to file offsets. This is problematic if the provided `IBinaryStreamReader` is in fact reading a PE in its mapped form (e.g. from a running process).

Suggestion is to extend the `PEFile` API to also support files in mapped mode.

Washi1337:
Implemented using 2c7bbd3

PR #96, opened 2020-09-20
Handle zero-length sections

SapientGuardian:
This PR fixes https://github.com/Washi1337/AsmResolver/issues/95 by:

* Adding a test case which covers the scenario in which a section has a length of 0
* Allowing PhysicalContents to be null in VirtualSegment
* Creating a null PhysicalContents in PEFile.FromReader if applicable
* Maintaining the current fileOffset in PEFile.AlignSections
* Lifting RVA from PhysicalContents to VirtualSegment, while keeping the two in sync


Washi1337:
Great work, thanks again for spotting the issue! 👍 

Just as a heads up. The `FileOffset` property will be refactored soon to `Offset`, since a new feature of AsmResolver will allow you to read PE files from other data sources than files on the disk. This includes PEs that are mapped in memory (see #64 and feature/pefile-mapped-mode-reading).

Issue #132, opened 2021-01-31
Add C# 9.0 nullable annotations


Washi1337:
**Summary**
To prevent all kinds of issues and warnings with newer projects, we should move to C# 9.0 and annotate the API with nullable reference type annotations.

This is going to be quite a task, but probably worth it in the end.

**Projects that have been ported:**
- [x] AsmResolver
- [x] AsmResolver.PE.File
- [x] AsmResolver.PE
- [x] AsmResolver.PE.Win32Resources
- [x] AsmResolver.DotNet

**Challenges:**
- `LazyVariable<T>` is going to be interesting, since `T` is used for both reference types as well as value types. Changing it to `T?` would therefore not work out of the box, since they have different semantics for reference types as opposed to value types. It is also used everywhere throughout the solution, so excluding it might not work that well either.

PR #142, opened 2021-02-26
[WIP] Nullability annotations

moien007:
Regarding issue #132; I made a branch `nullable` that adds annotations and null checks.

### The progress:

* [x]  AsmResolver
* [X]  AsmResolver.PE.File
* [ ]  AsmResolver.PE
* [x]  AsmResolver.PE.Win32Resources
* [ ]  AsmResolver.DotNet

### Questions
- in https://github.com/moien007/AsmResolver/commit/5e2c13b9bd04314eccd703099d8b795968188e68 `Name` is changed to be `string.Empty` instead of `null`, would this count as breaking change?

### DIFF
--- a/src/AsmResolver.DotNet/AsmResolver.DotNet.csproj
+++ b/src/AsmResolver.DotNet/AsmResolver.DotNet.csproj
@@ -5,6 +5,7 @@
         <Description>High level .NET image models for the AsmResolver executable file inspection toolsuite.</Description>
         <PackageTags>exe pe directories imports exports resources dotnet cil inspection manipulation assembly disassembly</PackageTags>
         <TargetFramework>netstandard2.0</TargetFramework>
+        <Nullable>enable</Nullable>
         <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
         <NoWarn>1701;1702;NU5105</NoWarn>
         <LangVersion>9</LangVersion>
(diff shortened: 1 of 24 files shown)


Washi1337:
First of all, thanks for taking the initiative on this :) :+1: 

However, I can't merge this into `development` just yet. Enabling NRT on all projects without actually using NRT litters the entire solution with warnings. Just checking out the [AppVeyor build log](https://ci.appveyor.com/project/Washi1337/asmresolver/build/messages) shows that enabling it in all projects immediately results in close to 700 warnings alone. I would rather leave `master` and `development` out of that as much as possible. 

From an administrative standpoint, maybe we should treat this `nullable` branch as a separate "development" branch, which keeps pulling from the actual `development` branch and add the nullable annotations there. This allows for further development of the library without having to wait for everything to be annotated.

As for your question. Setting `PESection.Name` and `SectionHeader.Name` to `string.Empty` is an acceptable change. `PESection.Name` is not supposed to be `null`, as section names are always 8 bytes in the final file. Furthermore, `SectionHeader` is not really meant to be used outside of AsmResolver itself directly anyways.

moien007:
> From an administrative standpoint, maybe we should treat this nullable branch as a separate "development" branch, which keeps pulling from the actual development branch and add the nullable annotations there. This allows for further development of the library without having to wait for everything to be annotated.

Yes, that's why I created the `nullable` branch. The question is, is this up to you to create this on the upstream?

Washi1337:
Superseded by #185 #186 #188 #189 .

PR #145, opened 2021-03-02
Fixed the issue with writing GenericParams

Charterino:
Before this change, if you add more than 1 GenericParameter to a method definition, all of them would be written with the same number.

### DIFF
--- a/src/AsmResolver.DotNet/Cloning/MemberCloner.cs
+++ b/src/AsmResolver.DotNet/Cloning/MemberCloner.cs
@@ -385,8 +385,6 @@ namespace AsmResolver.DotNet.Cloning
         {
             var clonedParameter = new GenericParameter(parameter.Name, parameter.Attributes);
 
-            //clonedParameter.Number = parameter.Number;
-
             foreach (var constraint in parameter.Constraints)
                 clonedParameter.Constraints.Add(CloneGenericParameterConstraint(context, constraint));
 
--- a/src/AsmResolver.DotNet/Cloning/MemberCloner.cs
+++ b/src/AsmResolver.DotNet/Cloning/MemberCloner.cs
@@ -8,12 +8,12 @@ using AsmResolver.PE.DotNet.Metadata.Tables;
 namespace AsmResolver.DotNet.Cloning
 {
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/AsmResolver.DotNet/Cloning/MemberCloner.cs
```diff
@@ -385,7 +385,7 @@ private GenericParameter CloneGenericParameter(MemberCloneContext context, Gener
         {
             var clonedParameter = new GenericParameter(parameter.Name, parameter.Attributes);
 
-            clonedParameter.Number = parameter.Number;
+            //clonedParameter.Number = parameter.Number;
```
Washi1337, line 388: Remove comment.

src/AsmResolver.DotNet/Serialized/SerializedGenericParameter.cs
```diff
@@ -31,7 +31,7 @@ public SerializedGenericParameter(ModuleReaderContext context, MetadataToken tok
             _row = row;
 
             Attributes = _row.Attributes;
-            Number = row.Number;            
+            //Number = row.Number;
```
Washi1337, line 34: Remove comment

Washi1337 (changes requested):
There's a couple of leftover statements commented out. Other than that LGTM!

src/AsmResolver.DotNet/Cloning/MemberCloner.cs
Charterino: You can't set the number manually now.

src/AsmResolver.DotNet/Serialized/SerializedGenericParameter.cs
Charterino: You can't set the number manually now. Setting the number would be done through reordering the owner array.

src/AsmResolver.DotNet/Cloning/MemberCloner.cs
Charterino: Perhaps I should delete the line altogether, is that what you meant? 😛

Washi1337: Yes, remove the entire line is what I meant 👍

PR #147, opened 2021-03-08
[WIP] Feature: State ITypeSignatureVisitor

JPaja:
Feature that would allow us to write static visitor so we can avoid generating new visitor for each state


Washi1337 (approved):

Washi1337:
LGTM, thanks!

PR #154, opened 2021-03-13
Fix cloning ldarg operand

oSumAtrIX:
Before it could happen that it can't find the parameter index of the ldarg operand in case of the method having no parameters


Washi1337:
Please fork from the `development` branch and PR into `development` to prevent merge conflicts later. `master` stays at all time in sync with nuget :)

PR #170, opened 2021-04-20
Fix naming & xml docs and add PropertyAttributes.None

oSumAtrIX:
AsmResolver was missing the `None` attribute and I had to use 0 instead.


oSumAtrIX:
RtSpecialName has to be renamed to RuntimeSpecialName in the Solution in order to solve errors

Washi1337:
Thanks!

Issue #63, opened 2020-06-02
Read/write support for VTable fixups data directory


Washi1337:
**Summary:**
The .NET file format supports unmanaged exports of method definitions defined in the metadata tables by the means of vtable fixups. AsmResolver should support reading/writing as well:

<details>
<summary>Example ilasm file</summary>

```
.assembly Test {}

.vtfixup [1] int32 fromunmanaged at VT_01

.method public static void ManagedMethod()
{
   .vtentry 1:1 // entry 1, slot 1
   .export [1] as UnmanagedExportName // export 1, name UnmanagedExportName

   ldstr "Hello, world!"
   call void [mscorlib] System.Console::WriteLine(string)
   ret
}

.data VT_01 = int32(0)
```

</details>

**Additional info:**
- Section 15.5.1 (page 187) of ECMA-335.
- Section 25.3.3.3 (page 284) of ECMA-335.
- [Structure of IMAGE_COR_VTABLEFIXUP](https://github.com/dotnet/runtime/blob/ec6bec455514070cc69e4120974796338bcb7cb7/src/coreclr/src/inc/corhdr.h#L1272)
- [VTable fixup constants](https://github.com/dotnet/runtime/blob/ec6bec455514070cc69e4120974796338bcb7cb7/src/coreclr/src/inc/corhdr.h#L161)
- [Implementation vtable fixups in runtime](https://github.com/dotnet/runtime/blob/ec6bec455514070cc69e4120974796338bcb7cb7/src/coreclr/src/vm/ceeload.cpp#L6078)
- Requires #62 for reading/writing to the exports directory of a PE image.

Washi1337:
Implemented in #173

PR #173, opened 2021-04-24
Read/write support for VTable fixups

Anonym0ose:
Closes https://github.com/Washi1337/AsmResolver/issues/63


src/AsmResolver.PE/DotNet/Builder/DotNetSegmentBuffer.cs
```diff
@@ -30,6 +27,11 @@ public DotNetSegmentBuffer(IDotNetDirectory dotNetDirectory)
             AddIfPresent(dotNetDirectory.DotNetResources);
             AddIfPresent(dotNetDirectory.StrongName);
             AddIfPresent(dotNetDirectory.VTableFixups);
+            if (dotNetDirectory.VTableFixups?.Count > 0)
+            {
+                var buffer = new VTableEntriesBuffer(dotNetDirectory.VTableFixups);
+                AddIfPresent(buffer);
```
Washi1337, line 33: Unnecessary AddIfPresent call. `buffer` is always non-null. We can just use `_segments.Add` here.

Washi1337 (changes requested):
Would probably also need some unit tests to ensure that it works now and in the future.

src/AsmResolver.PE/DotNet/VTableFixups/VTableFixupDirectory.cs
```diff
@@ -0,0 +1,71 @@
...
+using AsmResolver.IO;
+
+namespace AsmResolver.PE.DotNet.VTableFixups
+{
+    /// <summary>
+    /// Represents the VTable Fixup Directory in the Cor20 header.
+    /// </summary>
+    public class VTableFixupDirectory : Collection<VTableFixup>, ISegment
+    {
+        private readonly Dictionary<VTableFixup, uint> _tableRvas = new();
```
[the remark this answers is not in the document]
Anonym0ose: The purpose of the dictionary is to store the location of the entries of the table, and the RVA is actually the first field of the structure of VTable Fixups: https://github.com/dotnet/runtime/blob/ec6bec455514070cc69e4120974796338bcb7cb7/src/coreclr/src/inc/corhdr.h#L1272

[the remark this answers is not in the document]
Washi1337: Yes, but this setup forces every implementation of `IPEFileBuilder` to place the actual lists of tokens right after the vtablefixups directory. Spec states that the actual lists of metadata tokens can be _anywhere_ in the PE, as long as it's in a section with read/write access flags (ECMA-335, Section II.25.3.3.3, page 284). Therefore, I think a different `IPEFileBuilder` implementation (such as one that preserves as much as possible of the original PE file) should be able to decide a different location for these vtables.

Perhaps we can let `VTableFixup` implement `ISegment` or inherit from `SegmentBase` instead so that it becomes relocatable by a `IPEFileBuilder` instance?

src/AsmResolver.PE/DotNet/VTableFixups/Builder/VTableFixupsDirectoryBuffer.cs
```diff
@@ -0,0 +1,56 @@
+﻿using AsmResolver.IO;
+
+namespace AsmResolver.PE.DotNet.VTableFixups.Builder
+{
+    /// <summary>
+    /// Provides a mechanism for building the VTable fixups directory in a PE file.
+    /// </summary>
+    public class VTableFixupsDirectoryBuffer : ISegment
+    {
+        private readonly SegmentBuilder _vtableDirectory = new();
```
Washi1337, line 10: Unnecessary `SegmentBuilder`, since only a single vtable fixups directory is added. Replace with the actual vtable directory?

src/AsmResolver.PE/DotNet/VTableFixups/VTableFixup.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            get;
+        } = new ();
+
+
+        /// <summary>
+        /// Creates a new VTable fixup.
+        /// </summary>
+        /// <param name="type"></param>
+        public VTableFixup(VTableType type) => Tokens.Type = type;
```
Washi1337, line 24: Constructor should be declared before all properties for consistency.

src/AsmResolver.PE/DotNet/VTableFixups/Builder/VTableFixupsDirectoryBuffer.cs
```diff
@@ -0,0 +1,56 @@
...
+        public uint Rva => _vtableDirectory.Rva;
+
+        /// <inheritdoc />
+        public bool CanUpdateOffsets => true;
+
+        /// <summary>
+        /// Creates a new VTable fixups directory buffer.
+        /// </summary>
+        /// <param name="vtableFixupsDirectory"></param>
+        public VTableFixupsDirectoryBuffer(VTableFixupsDirectory vtableFixupsDirectory)
```
Washi1337, line 26: Constructor should be declared before all properties for consistency.

Washi1337 (changes requested):
Few minor things, but other than that LGTM!

Washi1337:
Awesome! Thanks a lot!

Washi1337 (approved):

PR #174, opened 2021-04-25
IFileService interfaces

### LINKED ISSUE #115: Read PE via stream/buffer
Can we get a way to read a PE File via stream/buffer without reading all bytes at once? Some PE files are ~400mb, its insane to read them all into memory when we're looking into the PE header only.

Washi1337:
Adds `IFileService` that is used to abstract away the way files are opened and read from the disk. This allows for reading files using either simple byte arrays, or more advanced memory mapped files.

Closes #115.

### DIFF
--- a/docs/dotnet/basics.rst
+++ b/docs/dotnet/basics.rst
@@ -30,17 +30,17 @@ Opening a .NET module can be done through one of the `FromXXX` methods from the 
 
 .. code-block:: csharp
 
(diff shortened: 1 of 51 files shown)


Washi1337:
Problem here is that `AssemblyResolverBase` still reads the assembly using `FromFile` which doesn't use memory mapped I/O.

PR #191, opened 2021-07-26
Fix stack push/pop count when encountering a method with a modreq return type

zsr2531:
This PR fixes an issue with the `GetStackPopCount` and `GetStackPushCount` methods where if the return type is a `CustomModifierTypeSignature`, whose `BaseType` is `void`, it would get treated as if it actually returned a value.


src/AsmResolver.DotNet/Code/Cil/CilInstructionExtensions.cs
```diff
@@ -23,8 +24,8 @@ public static class CilInstructionExtensions
         public static int GetStackPopCount(this CilInstruction instruction, CilMethodBody? parent)
         {
             return GetStackPopCount(instruction,
-                parent is null
-                || !(parent.Owner.Signature?.ReturnsValue).GetValueOrDefault());
+                parent == null
+                || HasReturnValue(parent.Owner.Signature));
```
Washi1337, line 28: Possible NRE when the `Signature` property is `null` (as a result of invalid metadata).

src/AsmResolver.DotNet/Code/Cil/CilInstructionExtensions.cs
```diff
@@ -242,5 +243,13 @@ public static Parameter GetParameter(this CilInstruction instruction, ParameterC
             }
         }
 
+        private static bool HasReturnValue(MethodSignature signature)
+        {
+            var ret = signature.ReturnType;
+            while (ret is CustomModifierTypeSignature custom)
+                ret = custom.BaseType;
+
+            return ret?.IsTypeOf("System", "Void") ?? false;
```
Washi1337, line 252: Unnecessary null-coalescing.

Washi1337 (changes requested):

Washi1337:
Actually, I just realized that `MethodSignatureBase` already defines a property `ReturnsValue`. We should probably move the impl of `HasReturnValue` to that property instead.

zsr2531:
Closing because I'll open another PR that is based on the correct branch.

PR #210, opened 2021-10-03
Bump test app framework version

JPaja:
Due to new visual studio installer not having .net 4.0 install option, i bumped test app versions to later framework version.


Washi1337:
Thanks!

PR #211, opened 2021-10-03
[Fix] Method analyzer doesn't locate override implementations when layered

### LINKED ISSUE #198: Method analyzer doesn't locate override implementations when layered
## Describe the bug
Using feature/workspaces DotNetWorkspace to analyze an abstract method should find all of the direct implementations, but only looks for types that implement the abstract method's declaring type.

## To Reproduce
```
using System;
using System.Linq;
using AsmResolver.DotNet;
using AsmResolver.Workspaces.DotNet;

var path = "Issue.dll";
var module = ModuleDefinition.FromFile(path);

var workspace = new DotNetWorkspace();
workspace.Assemblies.Add(module.Assembly);
workspace.Analyze();

var type = module.TopLevelTypes.First(d=>d.Name == "TestAbstractClass");
var method = type.Methods.First(d => d.Name == "TestAbstractBool");
            
var node = workspace.Index.GetOrCreateNode(method);
var impls = node.BackwardRelations.GetObjects(DotNetRelations.ImplementationMethod).ToArray();
foreach (var impl in impls)
{
   Console.WriteLine(impl.DeclaringType.Name);
}
Console.WriteLine(impls.Length == 3);
```
[Issue.zip](https://github.com/Washi1337/AsmResolver/files/7101450/Issue.zip)

## Proposal
Should perhaps look recursively for the types that implement the abstract method's declaring type

## Platform
 - OS: Linux
 - AsmResolver Version: **feature/workspaces** branch

JPaja:
Fixes #198

### DIFF
--- a/src/AsmResolver.Workspaces.DotNet/Analyzers/Definition/TypeAnalyzer.cs
+++ b/src/AsmResolver.Workspaces.DotNet/Analyzers/Definition/TypeAnalyzer.cs
@@ -54,11 +54,13 @@ namespace AsmResolver.Workspaces.DotNet.Analyzers.Definition
             var index = context.Workspace.Index;
             var node = index.GetOrCreateNode(subject);
 
-            // Register base type relation.
-            if (subject.BaseType is { } baseType)
+            // Register base type relations.
+            var baseType = subject.BaseType;
+            while(baseType is not null)
             {
                 var relatedNode = index.GetOrCreateNode(baseType);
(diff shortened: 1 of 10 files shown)


Washi1337:
This is unfortunately not an acceptable solution, since it completely flattens (= destroys) the type hierarchy that is supposed to be encoded in the dependency graph. This is important if we want to query information like the immediate inheritor or the immediate base definition of a member.

Detection like this should happen at the detection level, not in the graph itself.

PR #212, opened 2021-10-03
Adjusted workspaces for NRT update

JPaja:


Washi1337:
I think we should move the NRT annotations for `GenericContext` into the dev branch, as this is a feature that does not necessarily contribute directly to the workspaces feature, but is very useful nonetheless. We can maybe squeeze it into the 4.7.1 hotfix. Would you mind making a separate PR for this? :)

PR #223, opened 2021-11-27
Fix: Multiple generics support in TypeNameParser

### LINKED ISSUE #222: CustomAtribute TypeSignature nested generic type parse error
## Describe the bug
On custom argument 
```csharp
 [TypeConverter(typeof(StringBasedTypeConverter<Transform, Transform.Parser>))]
 ```
on its serialized string representation
```csharp
PaintDotNet.Markup.StringBasedTypeConverter`2[PaintDotNet.UI.Media.Transform,PaintDotNet.UI.Media.Transform+Parser], PaintDotNet.Base, Version=4.302.7947.34042, Culture=neutral, PublicKeyToken=null
```
it breaks when it gets to `+` character since its expecting closed bracket or comma.

From my observation `+` character should handle nested types

## To Reproduce
```csharp
var asm = AssemblyDefinition.FromFile(@"PaintDotNet.Framework.dll");
var type = (TypeDefinition)asm.ManifestModule.LookupMember(0x0200008B); //PaintDotNet.UI.Media.Transform
var customAttribute = type.CustomAttributes[0];
Console.WriteLine(customAttribute.Signature);
```

## Expected behavior
To print to standard output

## Platform
 - OS: [Windows 10]
 - AsmResolver Version: [e.g. 4.8-dev]
 
 ## Files
PaintDotNet.Framework.dll from Paint.Net project
https://www.getpaint.net/download.html

JPaja:
Fixes #222

### DIFF
--- a/src/AsmResolver.DotNet/Signatures/Types/Parsing/TypeNameParser.cs
+++ b/src/AsmResolver.DotNet/Signatures/Types/Parsing/TypeNameParser.cs
@@ -248,7 +248,7 @@ namespace AsmResolver.DotNet.Signatures.Types.Parsing
         private TypeSignature ParseGenericTypeArgument(GenericInstanceTypeSignature genericInstance)
         {
             var extraBracketToken = TryExpect(TypeNameTerminal.OpenBracket);
-            var result = ParseSimpleTypeSpec();
+            var result = ParseTypeSpec();
             if (extraBracketToken.HasValue)
                 Expect(TypeNameTerminal.CloseBracket);
             return result;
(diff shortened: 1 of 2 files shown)


JPaja:
Ok so my conclusion from this is
when generics are just in 1 braces they cannot have assembly info
[Generic1, Generic2]
but when its double then it can have commas inside
[[Generic, Assembly, Version=123], [Generic2, Assembly2, Version=333]]

Washi1337:
Thanks, good catch!

PR #225, opened 2021-11-27
Fix: Throw on generic context GetTypeArgument + Add: GenericContext From  methods

JPaja:
In case when we want to partially replace generic values (only Vars, or only MVars) on methods that contains both, we don't want to throw error when second provider is missing

### DIFF
--- a/src/AsmResolver.DotNet/Signatures/GenericContext.cs
+++ b/src/AsmResolver.DotNet/Signatures/GenericContext.cs
@@ -51,6 +51,7 @@ namespace AsmResolver.DotNet.Signatures
 
         /// <summary>
         /// Resolves a type parameter to a type argument, based on the current generic context.
+        /// If generic argument provider is not initialised for type parameter, than it will return itself.
         /// </summary>
         /// <param name="parameter">The parameter to get the argument value for.</param>
         /// <returns>The argument type.</returns>
--- a/src/AsmResolver.DotNet/Signatures/GenericContext.cs
+++ b/src/AsmResolver.DotNet/Signatures/GenericContext.cs
@@ -99,20 +99,5 @@ namespace AsmResolver.DotNet.Signatures
                 (TypeSpecification typeSpec, _) => FromTypeSpecification(typeSpec),
                 _ => null
             };
-
-        /// <summary>
-        /// Tries to get method and/or type generic context from <see cref="IMemberDescriptor"/>.
-        /// </summary>
-        /// <param name="member">Member to get generic context from.</param>
-        /// <returns>Generic context</returns>
-        public static GenericContext? FromMember(IMemberDescriptor member) =>
-            (member.DeclaringType, member) switch
-            {
-                (_, TypeSpecification typeSpec) => FromTypeSpecification(typeSpec),
-                (_, MethodSpecification methodSpec) => FromMethodSpecification(methodSpec),
-                (_, GenericInstanceTypeSignature typeSig) => new GenericContext(typeSig, null),
-                ({ } declaringType, _) => FromMember(declaringType),
-                _ => null
-            };
     }
 }
--- a/src/AsmResolver.DotNet/Signatures/GenericContext.cs
+++ b/src/AsmResolver.DotNet/Signatures/GenericContext.cs
@@ -1,5 +1,6 @@
 using System;
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using AsmResolver.DotNet.Signatures.Types;
 
 namespace AsmResolver.DotNet.Signatures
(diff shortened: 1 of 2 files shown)


src/AsmResolver.DotNet/Signatures/GenericContext.cs
```diff
@@ -18,39 +19,45 @@ public GenericContext(IGenericArgumentsProvider? type, IGenericArgumentsProvider
...
         public IGenericArgumentsProvider? Method
         {
             get;
         }
 
         /// <summary>
-        /// Enters a new generic context with a new type providing type arguments. 
+        /// Returns true if both Type and Method providers are null
+        /// </summary>
+        public bool IsEmpty => Type is null && Method is null;
```
Washi1337, line 42: Add `[MemberNotNullWhen]` attributes referencing both the `Type` and `Method` properties.

src/AsmResolver.DotNet/Signatures/GenericContext.cs
```diff
@@ -62,14 +69,111 @@ public TypeSignature GetTypeArgument(GenericParameterSignature parameter)
...
 
             if (parameter.Index >= 0 && parameter.Index < argumentProvider.TypeArguments.Count)
                 return argumentProvider.TypeArguments[parameter.Index];
 
             throw new ArgumentOutOfRangeException();
         }
+
+
+        /// <summary>
+        /// Gets type generic context from <see cref="TypeSpecification"/>.
```
Washi1337, line 84: Gets **a** type generic context from **a** TypeSpecification.

Applies to all `FromXXX` methods.

src/AsmResolver.DotNet/Signatures/GenericContext.cs
```diff
@@ -62,14 +69,111 @@ public TypeSignature GetTypeArgument(GenericParameterSignature parameter)
...
+        /// </summary>
+        /// <param name="type">Type specification to get generic context from.</param>
```
Washi1337, line 86: Type specification to get **the** generic context from.

Applies to all `FromXXX` methods.

src/AsmResolver.DotNet/Signatures/GenericContext.cs
```diff
@@ -62,14 +69,111 @@ public TypeSignature GetTypeArgument(GenericParameterSignature parameter)
...
+            _ => default
+        };
+
+        /// <summary>
+        /// Gets method and/or type generic context from <see cref="MethodSpecification"/>.
+        /// </summary>
+        /// <param name="method">Method specification to get generic context from.</param>
+        /// <returns>Generic context.</returns>
+        public static GenericContext FromMethod(MethodSpecification method) =>
+            (method.DeclaringType, method.Signature) switch
```
Washi1337, line 119: This is some pretty hefty `switch` pattern matching abuse. Maybe consider normal if statements for readability?

src/AsmResolver.DotNet/Signatures/GenericContext.cs
```diff
@@ -18,39 +19,45 @@ public GenericContext(IGenericArgumentsProvider? type, IGenericArgumentsProvider
...
-        /// Enters a new generic context with a new method providing type arguments. 
+        /// Enters a new generic context with a new method providing type arguments.
         /// </summary>
         /// <param name="method">The new method providing the type arguments.</param>
         /// <returns>The new generic context.</returns>
         public GenericContext WithMethod(IGenericArgumentsProvider method) => new GenericContext(Type, method);
 
         /// <summary>
         /// Resolves a type parameter to a type argument, based on the current generic context.
+        /// If generic argument provider is not initialised for type parameter, than it will return itself.
```
Washi1337, line 60: This sentence is not really clear.
- What is meant by _if it is not initialized for type parameter_? 
- The word _itself_ can now refer to the generic argument provider or to the type parameter.

I think you mean to say something in the lines of: 

> If a type parameter within the signature references a parameter that is not captured by the context (i.e. the corresponding generic argument provider is set to `null`), then this type parameter will not be substituted.


We might consider moving this to the `<remarks>` section as well.

test/AsmResolver.DotNet.Tests/Signatures/GenericContextTest.cs
```diff
@@ -72,38 +73,320 @@ public void ResolveMethodGenericParameterWithTypeAndMethod()
...
+            Assert.Equal(context, context2);
+            Assert.False(context.IsEmpty);
+            Assert.Equal(genericInstance, context.Type);
+            Assert.Null(context.Method);
+        }
+
+        [Fact]
+        public void ParseGenericFromNonGenericTypeSignature()
+        {
+            var type = new TypeDefinition("","Test type", TypeAttributes.Public);
```
Washi1337, line 131: Missing spaces in format. Few other places in this same file with the same problem.

Washi1337 (changes requested):
Few things but otherwise LGTM

src/AsmResolver.DotNet/Signatures/GenericContext.cs
JPaja: Not relevant since we don't know which member is not null

Washi1337 (approved):

PR #238, opened 2021-12-24
Fix/ AssemblyDefinition Write to relative path

ds5678:
Previously, `AssemblyDefinition.Write` would throw a `DirectoryNotFoundException` if provided a relative path without any directories, such as `output.dll`. This pull request make it use the full path instead, avoiding this issue.


Washi1337:
Thanks for the catch!

PR #239, opened 2021-12-27
Feature/add documentation for generic instance methods

ds5678:


Washi1337:
A good chunk of this is already covered in the article about type signatures (for generic types at least). 

https://asmresolver.readthedocs.io/en/latest/dotnet/type-signatures.html

I think it might be better to start a generic article on methods themselves, as there is a lot to talk about when it comes to creating them (e.g. MethodDefinition vs MethodSpecification, signatures and more).

Not saying everything has to be written right away, but this is I think a structure that would make more sense.

ds5678:
Hi Washi. I created this because it was the least intuitive thing in my transition from Cecil, and I wanted others to have it easier. 

I renamed the file to be less limiting.

JPaja:
I would recommend adding documentation on GenericContext (from dev branch) and InstantiateGenericTypes functions

Examples:
https://github.com/Washi1337/AsmResolver/blob/development/test/AsmResolver.DotNet.Tests/Signatures/GenericTypeActivatorTest.cs
https://github.com/Washi1337/AsmResolver/blob/development/test/AsmResolver.DotNet.Tests/Signatures/GenericContextTest.cs

ds5678:
Is there anything I can do for this pull request? I feel that I don't know enough to expand it

Washi1337:
My apologies I have been particularly busy lately that I completely forgot to come back to this PR. 

@ds5678 We don't have to add anything else topic-wise for this PR, but perhaps your code could be explained slightly more. From the text alone, it is unclear where the `corlibScope` variable comes from, and the naming convention used is slightly inconsistent (e.g. `arrayRef` vs `emptyMethodDefinition`).

Side note: Your example code can actually be simplified more by replacing the explicit instantiation of a `GenericInstanceTypeSignature` with

```csharp
var listOfInt32 = listTypeRef.MakeGenericInstanceType(typeFactory.Int32);
```

This makes me wonder if this should be stated more explicitly in https://github.com/Washi1337/AsmResolver/blob/master/docs/dotnet/type-signatures.rst#decorating-type-signatures. I realize now that I might have tucked this away a little too much under the shortcuts table at the very end of this article.

Additionally, your second sample makes me wonder whether we should introduce a `MakeGenericInstanceMethod` extension method as well to automatically create `GenericInstanceMethodSignature` and/or `MethodSpecification` instances. This does not have to be part of this specific PR either.

@JPaja `GenericContext` probably deserves its own article, I don't think it would fit on a page that is purely about creating method definitions and/or references.

ds5678:
@Washi1337 is this more in line with what you had in mind?

Washi1337:
Yes that is better! Thanks a lot. After readthedocs has realized the commit has merged and recompiled the HTML, it should be online.

PR #244, opened 2022-01-19
Feature/ FromValue methods for DataBlobSignature and Constant

ds5678:
This pull request adds FromValue static methods to DataBlobSignature and Constant.

### DIFF
--- a/src/AsmResolver.DotNet/Constant.cs
+++ b/src/AsmResolver.DotNet/Constant.cs
@@ -81,5 +81,161 @@ namespace AsmResolver.DotNet
         /// This method is called upon initialization of the <see cref="Value"/> property.
         /// </remarks>
         protected virtual DataBlobSignature? GetValue() => null;
+
+        /// <summary>
+        /// Create a <see cref="Constant"/> from a value
+        /// </summary>
+        /// <param name="value">The value to be assigned to the constant</param>
+        /// <returns>
+        /// A new <see cref="Constant"/> with the correct <see cref="Type"/> and <see cref="Value"/>
+        /// </returns>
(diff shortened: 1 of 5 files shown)


Washi1337:
Great suggestion! My only point of critique here is that your tests are either named incorrectly, or they should only test the `DataBlobSignature.FromValue` instead of `Constant.FromValue`. I think both are fine, but let's be consistent :)

ds5678:
How would you suggest I rename them?

Washi1337:
My suggestion would be to either move the tests to the `ConstantTest` class, as your tests are mostly testing for the correct workings of `Constant.FromValue`, or change the `Constant.FromValue` calls to `DataBlobSignature.FromValue` so that they can stay in the `DataBlobSignatureTest` file.

ds5678:
I originally thought about putting there, but I felt that they bloated the test class and took away from the other tests there. How do you feel about me making `ConstantTest` into a partial class with the FromValue tests in a file called `ConstantTest.FromValue.cs`?

ds5678:
If you feel differently about the bloatedness, it doesn't bother me that much to dump them all into `ConstantTest.cs`.

ds5678:
Whichever you prefer

Washi1337:
> I originally thought about putting there, but I felt that they bloated the test class and took away from the other tests there. How do you feel about me making `ConstantTest` into a partial class with the FromValue tests in a file called `ConstantTest.FromValue.cs`?

That is a good point, yes I think this is the right solution. If you can do that I'll merge this PR.

ds5678:
Done!

Washi1337:
Perfect. Thanks again!

PR #248, opened 2022-02-11
[Workspaces] Bug fixes

JPaja:
Fixes:
* Linking Type/Method specification to its definition
* Throwing when MethodBody is not IL
* Avoided resolving of reference if Parent/Scope is not in workspace.Assemblies

Requres #247


src/AsmResolver.Workspaces.DotNet/Analyzers/Definition/MethodAnalyzer.cs
```diff
@@ -46,7 +46,7 @@ private static void ScheduleMembersForAnalysis(AnalysisContext context, MethodDe
             }
 
             // Schedule method body for analysis.
-            if (subject.CilMethodBody is not null && context.HasAnalyzers(typeof(CilMethodBody)))
+            if (subject.IsIL && subject.CilMethodBody is not null && context.HasAnalyzers(typeof(CilMethodBody)))
```
Washi1337, line 49: Why is this extra check required? If the method body is a CIL body, then `CilMethodBody` is not `null`.

src/AsmResolver.Workspaces.DotNet/DotNetRelations.cs
```diff
@@ -47,13 +47,28 @@ public static class DotNetRelations
             "ReferenceMember",
             new Guid("ce11d2f6-a423-429d-ad37-2f073fdf63be"));
 
+        /// <summary>
+        /// Describes the relationship between a method definition and its specification.
+        /// </summary>
+        public static readonly ObjectRelation<MethodDefinition, MethodSpecification> ReferenceMethodSpecification = new(
```
Washi1337, line 53: Shouldn't this be a `IMethodDefOrRef` to `MethodSpecification`,  instead of `MethodDefinition` to `MethodSpecification`? A `MethodSpecification`  can also instantiate member references.

src/AsmResolver.Workspaces.DotNet/DotNetRelations.cs
```diff
@@ -47,13 +47,28 @@ public static class DotNetRelations
...
         /// Describes the relationship between a type definition and its reference.
         /// </summary>
         public static readonly ObjectRelation<TypeDefinition, TypeReference> ReferenceType = new(
             "ReferenceType",
             new Guid("3cc86779-338c-4165-a00c-da547a2e8549"));
 
+        /// <summary>
+        /// Describes the relationship between a type definition and its specification.
+        /// </summary>
+        public static readonly ObjectRelation<TypeDefinition, TypeSpecification> ReferenceTypeSpecification = new(
```
Washi1337, line 68: I don't understand the purpose of this relation. `TypeSpecification`s assign tokens to `TypeSignature`s. TypeSignatures can contain references to other types (e.g. `GenericType<TypeParam1, TypeParam2>` contains 3 types). Furthermore, these types within the signatures can be definitions as well as references. Reducing it to just type definitions flattens the dependency graph, no?

src/AsmResolver.Workspaces.DotNet/Analyzers/Definition/TypeSpecificationAnalyzer.cs
```diff
@@ -21,6 +21,13 @@ protected override void Analyze(AnalysisContext context, TypeSpecification subje
             {
                 context.ScheduleForAnalysis(subject.DeclaringType);
             }
+
+            if (subject.Resolve() is { } definition)
```
Washi1337, line 25: Why are we resolving here? Doesn't this flatten the dependency graph by skipping any type references?

Washi1337 (changes requested):

src/AsmResolver.Workspaces.DotNet/DotNetRelations.cs
Washi1337: Furthermore, they are also used for other decorator type sigs like `int[][][]` or `int[]&`. How does this fit then in this relation?

src/AsmResolver.Workspaces.DotNet/Analyzers/Definition/MethodAnalyzer.cs
JPaja: CilMethodBody can throw exception when body is native

src/AsmResolver.Workspaces.DotNet/DotNetRelations.cs
JPaja: This relation is later used in pr #249 for linking/chaining base types when they are type specification
in src/AsmResolver.Workspaces.DotNet/Analyzers/Implementation/ImplementationUtilities.cs
internal static IEnumerable<ITypeDefOrRef> GetBaseTypes(WorkspaceIndexNode baseNode)

JPaja: Maybe it should be named differently, for example ReferenceBaseTypeSpecification  for chaining TypeSpecification and its TypeDefOrRef generated from signature.

Or to do this with each TypeSignature , but only the first one (skipping the nested ones) for less complexity

src/AsmResolver.Workspaces.DotNet/Analyzers/Definition/MethodAnalyzer.cs
Washi1337: This sounds like a bug to me then. We'll leave it in this PR for now.

src/AsmResolver.Workspaces.DotNet/DotNetRelations.cs
Washi1337: Why not just `TypeDefinition` to `ITypeDefOrRef`, and reuse the previously defined relation?

src/AsmResolver.Workspaces.DotNet/DotNetRelations.cs
```diff
@@ -48,9 +48,17 @@ public static class DotNetRelations
             new Guid("ce11d2f6-a423-429d-ad37-2f073fdf63be"));
 
         /// <summary>
-        /// Describes the relationship between a type definition and its reference.
+        /// Describes the relationship between a method and its specification.
         /// </summary>
-        public static readonly ObjectRelation<TypeDefinition, TypeReference> ReferenceType = new(
+        public static readonly ObjectRelation<IMethodDefOrRef, MethodSpecification> ReferenceMethodSpecification = new(
```
Washi1337, line 53: Do we really need this? What about replacing the `MethodSpecification` with `IMemberDescriptor`, and then just reuse `ReferenceMember`, similar to how we reused the `ReferenceType` for type specs?

src/AsmResolver.Workspaces.DotNet/Analyzers/Reference/ExportedTypeAnalyzer.cs
```diff
@@ -18,6 +22,11 @@ protected override void Analyze(AnalysisContext context, ExportedType subject)
             if (context.Workspace is not DotNetWorkspace workspace)
                 return;
 
+            if (subject.Scope?.GetAssembly() is not { } assembly)
```
Washi1337, line 25: This code is repeated a lot in this PR. Extract function?

src/AsmResolver.Workspaces.DotNet/Analyzers/Reference/TypeSpecificationAnalyzer.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            if (subject.Scope?.GetAssembly() is not { } assembly)
+                return;
+            if (!workspace.Assemblies.Any(a => _comparer.Equals(a, assembly)))
+                return;
+
+            var definition = subject.Resolve();
+            if (definition is not { Module: { Assembly: { } } })
+                return;
+            if (!workspace.Assemblies.Contains(definition.Module.Assembly))
```
Washi1337, line 40: `definition.Module.Assembly` does not have to be repeated if we capture it in the pattern above.

src/AsmResolver.Workspaces.DotNet/DotNetRelations.cs
JPaja: Yes but in that case we loose support for FieldReferences, and than we would have to create new relation for those.

Washi1337: That would be the case if we used `IMethodDescriptor`, but we can use `IMemberDescriptor` instead to capture both.

Washi1337 (approved):

PR #258, opened 2022-02-24
Feature/Dependabot

ds5678:
This pull request:
* Bumps the versions on all outdated nuget dependencies
* Adds a script for automated dependency checking on the development branch by Dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "nuget" # See documentation for possible values
+    target-branch: "development"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
+
+  - package-ecosystem: "github-actions"
+    target-branch: "development"
+    # Workflow files stored in the
+    # default location of `.github/workflows`
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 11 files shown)


Washi1337:
While upgrading packages used by test projects is fine, we have to be careful with bumping nuget versions. We don't want to lose compatibility with systems that use older versions of this library. Note that the version used is a lower bound version number, not an upper bound. For this same reason, I actually have considered downgrading the dependency to `System.Text.Json`, or even removing it completely, and adding a simple implementation instead.

Would there be a way to configure dependabot to only upgrade projects within a certain sub directory?

ds5678:
I don't see how we could break other projects. AsmResolver targets net standard 2.0, so all of the packages it depends on also have a net standard target. Therefore any project that can use AsmResolver can also use all of its dependencies.

I'm on mobile, so I didn't check. However, I'm nearly certain that Dependabot can be restricted to subdirectories.

Speaking of targets, would you be open to a pr adding net 6 targets? I think they might have a small performance boost over the net standard target.

Washi1337:
> I don't see how we could break other projects. AsmResolver targets .NET Standard 2.0, so all of the packages it depends on also have a .NET Standard target. Therefore any project that can use AsmResolver can also use all of its dependencies.

This is a fair point. As long as the deps target .NET standard 2.0, it should be fine. Do you know if dependabot verifies this?

> Speaking of targets, would you be open to a PR adding .NET 6 targets? I think they might have a small performance boost over the .NET Standard target.

Would it even make sense to do this? As far as I know .NET standard is not an implementation, just a specification. Consumers that use .NET 6.0 will automatically use the implementation of this runtime. Any performance improvements will be introduced by the JIT or updates to the BCL.

ds5678:
I am not certain what all validation Dependabot does when checking for updates. However, in the worst possible scenario, we just reject a Dependabot pull request after the AppVeyor continuous integration fails.

I will investigate any possible .NET 6 performance disparities and come back with concrete numbers.

Washi1337:
> I am not certain what all validation Dependabot does when checking for updates. However, in the worst possible scenario, we just reject a Dependabot pull request after the AppVeyor continuous integration fails.

This seems like a strategy that will work.

Thanks again for setting it up.

PR #261, opened 2022-03-04
CilAssembler diagnostics performance regression

### LINKED ISSUE #260: Don't precompute (or allow disabling of) diagnostic prefix for CilAssembler
## Problem Description
I'm currently in the process of switching to AsmResolver from Cecil for generating managed assemblies, and I've found that AsmResolver is quite a bit slower than Cecil when it comes to saving assemblies (among other things, but this is what I'm investigating right now). 

Having now run it through a profiler (specifically dotTrace), I've found that approximately 12.5% of the total time to save assemblies is coming from the call to SafeToString in CilMethodBodySerializer#BuildRawCodeStream, the result of which is only used as the diagnostic prefix for CilAssembler should something go drastically wrong (which it hasn't yet, in my experimentation). 

Being able to disable this (or even better, compute it only if an error actually occurs) would be nice for applications where speed is preferable to easily reproducible log output - if need be, an option can be provided by the parent application to re-enable the logging for better error messages.

## Proposal
The addition of a feature flag/configuration option to disable certain calls such as this one which are only for ease of diagnostic output

## Alternatives
A better (i.e. more convenient from a developer point of view) however more complex to implement alternative would be the ability for certain time-intensive computations like this to be done as-needed instead of ahead of time.

## Additional context
N/A

Washi1337:
Adds constructor to `CilAssembler` that accepts a delegate for lazy loading the diagnostic prefix, in order to avoid expensive string formatting of method full names.

Closes #260


Washi1337:
Initial testing seems to indicate a great reduction in allocations and an performance improvement of around 10-20% on a binary with 10k methods with the signature:
```
System.Void *(System.Int32, System.Collections.Generic.List<object>, System.String[])
```

Old:

|                Method |          Mean |        Error |       StdDev |        Median |      Gen 0 |     Gen 1 |     Gen 2 |   Allocated |
|---------------------- |--------------:|-------------:|-------------:|--------------:|-----------:|----------:|----------:|------------:|
|       HelloWorld_Read |     165.79 us |     0.669 us |     0.559 us |     165.79 us |    10.0098 |    0.7324 |         - |    42.09 KB |
|  HelloWorld_ReadWrite |   7,697.47 us |   115.884 us |   108.398 us |   7,665.42 us |          - |         - |         - |  4601.72 KB |
|          CrackMe_Read |      45.22 us |     0.195 us |     0.182 us |      45.18 us |    10.0098 |         - |         - |    41.13 KB |
|     CrackMe_ReadWrite |  13,042.17 us |   565.130 us | 1,518.185 us |  12,444.93 us |  2000.0000 | 1000.0000 | 1000.0000 | 15072.52 KB |
|      ManyMethods_Read |     166.32 us |     2.026 us |     1.895 us |     165.73 us |    10.0098 |    0.4883 |         - |    42.09 KB |
| ManyMethods_ReadWrite | 222,102.96 us | 3,690.879 us | 3,271.868 us | 222,930.30 us | 15000.0000 | 6000.0000 | 2000.0000 | 89589.71 KB |

New:

|                Method |          Mean |        Error |       StdDev |        Median |      Gen 0 |     Gen 1 |     Gen 2 |   Allocated |
|---------------------- |--------------:|-------------:|-------------:|--------------:|-----------:|----------:|----------:|------------:|
|       HelloWorld_Read |     163.56 us |     0.608 us |     0.568 us |     163.53 us |    10.0098 |    0.7324 |         - |    42.09 KB |
|  HelloWorld_ReadWrite |   5,260.23 us |   674.524 us | 1,967.618 us |   4,081.19 us |          - |         - |         - |  4600.01 KB |
|          CrackMe_Read |      45.15 us |     0.151 us |     0.141 us |      45.12 us |    10.0708 |    0.0610 |         - |    41.13 KB |
|     CrackMe_ReadWrite |  12,686.78 us |   370.877 us |   977.038 us |  12,309.37 us |  1000.0000 | 1000.0000 | 1000.0000 | 15071.14 KB |
|      ManyMethods_Read |     165.71 us |     2.029 us |     1.898 us |     166.27 us |    10.0098 |    0.7324 |         - |    42.09 KB |
| ManyMethods_ReadWrite | 173,542.46 us | 3,127.802 us | 2,925.748 us | 172,520.52 us | 11000.0000 | 4000.0000 | 2000.0000 | 67732.45 KB |

PR #270, opened 2022-03-16
Use ConcurrentDictionary as backing cache for DefaultMetadataResolver

### LINKED ISSUE #269: DefaultMetadataResolver is not thread safe
## Describe the bug
Attempting to call TypeReference.Resolve() (or any property which requires a resolved reference, such as IsValueType) from multiple threads concurrently causes an exception:

```
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
   at System.Collections.Generic.Dictionary`2.FindValue(TKey key)
   at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
   at AsmResolver.DotNet.DefaultMetadataResolver.LookupInCache(ITypeDescriptor type)
   at AsmResolver.DotNet.DefaultMetadataResolver.ResolveTypeReference(TypeReference reference)
   at AsmResolver.DotNet.DefaultMetadataResolver.ResolveType(ITypeDescriptor type)
   at AsmResolver.DotNet.TypeReference.Resolve()
   at AsmResolver.DotNet.TypeReference.get_IsValueType()
   at AsmResolver.DotNet.TypeReference.ToTypeSignature()
--snip--
```

## To Reproduce
Create a bunch of type references, then convert them to type signatures, from many threads, e.g. using Parallel.ForEach. 

## Expected behavior
No exception was thrown, as ConcurrentDictionary or a lock was used.

## Screenshots
N/A - stacktrace provided above.

## Platform
 - OS: Windows 11
 - AsmResolver Version: 4.9.0

**Additional context**
As said above, a simple fix to this problem would probably be to switch DefaultMetadataResolver to use a ConcurrentDictionary as its backing cache.

SamboyCoding:
As discussed, closes #269. Some of the unit tests aren't passing on my local machine (41 of the 44 under Builder/TokenPreservation) but they don't pass on master either (but they seem to work on AppVeyor?) so I'm assuming I haven't broken anything. The error is 
`
System.BadImageFormatException: Input PE image does not contain a .NET metadata directory.
`.

### DIFF
--- a/src/AsmResolver.DotNet/DefaultMetadataResolver.cs
+++ b/src/AsmResolver.DotNet/DefaultMetadataResolver.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using AsmResolver.DotNet.Signatures;
 using AsmResolver.DotNet.Signatures.Types;
@@ -24,7 +25,7 @@ namespace AsmResolver.DotNet
         public DefaultMetadataResolver(IAssemblyResolver assemblyResolver)
         {
             AssemblyResolver = assemblyResolver ?? throw new ArgumentNullException(nameof(assemblyResolver));
-            _typeCache = new Dictionary<ITypeDescriptor, TypeDefinition>();
+            _typeCache = new ConcurrentDictionary<ITypeDescriptor, TypeDefinition>();
         }
 
         /// <inheritdoc />


ds5678:
> Some of the unit tests aren't passing on my local machine (41 of the 44 under Builder/TokenPreservation) but they don't pass on master either

All the tests pass on my machine, so maybe it just randomly failed for some reason.

SamboyCoding:
It's failing every time with the same error, no matter if I run from the IDE or through `dotnet test`.

Washi1337:
One reason for why these unit tests may be failing is that they require multiple versions of .NET to be installed on your computer. AsmResolver tests rebuilding of assemblies targeting different versions of .NET and .NET Core, partially by saving them to a temp directory, running them and testing for output. If you don't have the right .NET Core installed, that will probably fail.

Washi1337:
As long as AppVeyor is happy, it should be good! Thanks again for spotting and fixing it.

PR #278, opened 2022-03-23
Use actual types for private readonly fields

### LINKED ISSUE #271: [Question]: Why are interfaces used for private fields?
I noticed that some places in the code use an interface as the field type for private fields. One example location is here:

https://github.com/Washi1337/AsmResolver/blob/7cb4ae9b547a9a638c21107118239249de6ae5bf/src/AsmResolver.DotNet/DefaultMetadataResolver.cs#L14

As I understand things, interface calls are even less performant than virtual calls, so in this example, using a `Dictionary` or `ConcurrentDictionary` would improve performance. Making changes like this across the whole project could be a significant difference.

Is this something that you're open to?

ds5678:
Closes #271 

A note on the change from:
```cs
_typeCache.Remove(type);
```
To:
```cs
_typeCache.TryRemove(type, out _);
```
The runtime uses this behavior too:

https://github.com/dotnet/runtime/blob/3ae87395f638a533f37b8e3385f6d3f199a72f4f/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L1472

### DIFF
--- a/src/AsmResolver.DotNet/Builder/Metadata/Blob/BlobStreamBuffer.cs
+++ b/src/AsmResolver.DotNet/Builder/Metadata/Blob/BlobStreamBuffer.cs
@@ -14,7 +14,7 @@ namespace AsmResolver.DotNet.Builder.Metadata.Blob
     public class BlobStreamBuffer : IMetadataStreamBuffer
     {
         private readonly MemoryStream _rawStream = new();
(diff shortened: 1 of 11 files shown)


Washi1337 (approved):

Washi1337:
Thanks a lot!

ds5678:
You're welcome!

PR #279, opened 2022-03-23
.NET Bundle Manifests (AppHost / SingleFileHost)

### LINKED ISSUE #246: Support for .NET bundled / self-contained applications
## Problem Description
Latest .NET compilers (.NET Core 3.1 and up) can produce self-contained applications that statically link multiple .NET modules and its dependencies together into a single native binary. AsmResolver currently can only interpret this as a native binary, and individual modules need to be extracted manually.

## Proposal
Add support for reading and writing self-contained .NET executables.

## Alternatives
It could be argued whether this is a task that AsmResolver should solve, or an end-user of AsmResolver instead. However, it is expected that more applications will be using this model in the future, which may warrant first-class support for it regardless.

## Additional context
- https://docs.microsoft.com/en-us/dotnet/core/deploying/single-file
- https://github.com/dotnet/runtime/tree/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/installer/managed/Microsoft.NET.HostModel/AppHost
- https://github.com/dotnet/runtime/tree/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/native/corehost/apphost

This issue may split up in smaller issues.

Washi1337:
Includes read/write support for .NET bundle files.
- [x] Read and extract files from bundle manifest.
- [x] Write new manifests.
- [x] Write new single file executables using template AppHosts.
- [x] Compression support for file format v6+.
- [x] Automatic bundle ID generation based on manifest contents.
- [x] Update headers for PE AppHost templates (e.g. for GUI bit).
- [x] Update win32 resources for PE AppHost templates (e.g. for executable icons).
- [x] Documentation
- [x] Review

Closes #246


ds5678:
Excited!!

PR #283, opened 2022-03-27
[Feature]: Custom ReferenceImporter in MemberCloner

ds5678:
A draft pull request for addressing #93

One of the major design decisions I made was to require that any custom importers inherit from `CloneContextAwareReferenceImporter` instead of `ReferenceImporter` or a hypothetical `IReferenceImporter` interface. I felt that this was the cleanest way to maintain correct references between cloned members.

### DIFF
--- a/src/AsmResolver.DotNet/Cloning/CloneContextAwareReferenceImporter.cs
+++ b/src/AsmResolver.DotNet/Cloning/CloneContextAwareReferenceImporter.cs
@@ -5,7 +5,10 @@ namespace AsmResolver.DotNet.Cloning
     /// </summary>
     public class CloneContextAwareReferenceImporter : ReferenceImporter
     {
-        private readonly MemberCloneContext _context;
+        /// <summary>
+        /// The working space for this member cloning procedure.
+        /// </summary>
+        protected readonly MemberCloneContext _context;
 
         /// <summary>
         /// Creates a new instance of the <see cref="CloneContextAwareReferenceImporter"/> class.
--- a/src/AsmResolver.DotNet/Cloning/CloneContextAwareReferenceImporter.cs
+++ b/src/AsmResolver.DotNet/Cloning/CloneContextAwareReferenceImporter.cs
@@ -5,10 +5,7 @@ namespace AsmResolver.DotNet.Cloning
     /// </summary>
     public class CloneContextAwareReferenceImporter : ReferenceImporter
     {
-        /// <summary>
-        /// The working space for this member cloning procedure.
-        /// </summary>
-        protected readonly MemberCloneContext _context;
+        private readonly MemberCloneContext _context;
 
         /// <summary>
         /// Creates a new instance of the <see cref="CloneContextAwareReferenceImporter"/> class.
@@ -19,6 +16,11 @@ namespace AsmResolver.DotNet.Cloning
         {
             _context = context;
         }
+
+        /// <summary>
+        /// The working space for this member cloning procedure.
+        /// </summary>
+        protected MemberCloneContext Context => _context;
 
         /// <inheritdoc />
         protected override ITypeDefOrRef ImportType(TypeDefinition type)
--- a/src/AsmResolver.DotNet/Cloning/MemberCloneContext.cs
+++ b/src/AsmResolver.DotNet/Cloning/MemberCloneContext.cs
@@ -12,9 +12,15 @@ namespace AsmResolver.DotNet.Cloning
         /// Creates a new instance of the <see cref="MemberCloneContext"/> class.
         /// </summary>
         /// <param name="module">The target module to copy the cloned members into.</param>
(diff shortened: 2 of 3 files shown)


Washi1337:
Yes, this might be the only way that I can see for now that will ensure proper references will be used, while also allowing consumers to customize the behavior of the importer.

Side note: It may be best to not change existing constructors by adding optional parameters, but rather add constructor overloads instead. Adding optional parameters to a method / constructor is a breaking change on ABI level, forcing consumers to recompile their projects that use this feature.

ds5678:
Fixed!

src/AsmResolver.DotNet/Cloning/CloneContextAwareReferenceImporter.cs
```diff
@@ -5,7 +5,10 @@ namespace AsmResolver.DotNet.Cloning
     /// </summary>
     public class CloneContextAwareReferenceImporter : ReferenceImporter
     {
-        private readonly MemberCloneContext _context;
+        /// <summary>
+        /// The working space for this member cloning procedure.
+        /// </summary>
+        protected readonly MemberCloneContext _context;
```
Washi1337, line 11: Why was this made `protected`?

src/AsmResolver.DotNet/Cloning/MemberCloneContext.cs
```diff
@@ -12,10 +12,18 @@ public class MemberCloneContext
...
+        /// Creates a new instance of the <see cref="MemberCloneContext"/> class.
+        /// </summary>
+        /// <param name="module">The target module to copy the cloned members into.</param>
+        /// <param name="importerInstantiator">The instantiator for creating the reference importer</param>
+        public MemberCloneContext(ModuleDefinition module,
+            Func<MemberCloneContext, CloneContextAwareReferenceImporter>? importerInstantiator)
         {
             Module = module ?? throw new ArgumentNullException(nameof(module));
-            Importer = new CloneContextAwareReferenceImporter(this);
+            Importer = importerInstantiator?.Invoke(this) ?? new CloneContextAwareReferenceImporter(this);
```
Washi1337, line 26: Would a better name be a `factory` over an `instantiator`?

src/AsmResolver.DotNet/Cloning/MemberCloneContext.cs
```diff
@@ -12,10 +12,18 @@ public class MemberCloneContext
...
-        public MemberCloneContext(ModuleDefinition module)
+        public MemberCloneContext(ModuleDefinition module) : this(module, null) { }
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="MemberCloneContext"/> class.
+        /// </summary>
+        /// <param name="module">The target module to copy the cloned members into.</param>
+        /// <param name="importerInstantiator">The instantiator for creating the reference importer</param>
+        public MemberCloneContext(ModuleDefinition module,
+            Func<MemberCloneContext, CloneContextAwareReferenceImporter>? importerInstantiator)
```
Washi1337, line 23: If we only accept `CloneContextAwareReferenceImporter` instances here, do you think we should change the type of `MemberCloneContext.Importer`, for consistency?

Washi1337 (commented):
Sorry for the delay in response. I have added a few minor questions / comments.

src/AsmResolver.DotNet/Cloning/CloneContextAwareReferenceImporter.cs
ds5678: Only because I thought you would want it to be protected. I didn't have any specific reason on my mind other than that inheritors might need/want access to that field.

src/AsmResolver.DotNet/Cloning/MemberCloneContext.cs
ds5678: I like factory better, too 👍

src/AsmResolver.DotNet/Cloning/MemberCloneContext.cs
ds5678: That would be best for consistency, though I think it is a binary breaking change.

Washi1337: True, let's leave it as it is then for now.

src/AsmResolver.DotNet/Cloning/CloneContextAwareReferenceImporter.cs
ds5678: Implemented

src/AsmResolver.DotNet/Cloning/MemberCloneContext.cs
ds5678: Done

ds5678:
Should be good

Washi1337 (approved):
LGTM! Thanks.

PR #285, opened 2022-03-28
[QOL]: Change line ending setting to CRLF

ds5678:
The editor config was set to LF, and all the files are CRLF, so I had to constantly fix my line endings before making commits.


Washi1337 (approved):
Thanks for spotting.

PR #287, opened 2022-03-31
[Feature]: Allow .NET 6 assembly trimming for PE.File, PE and Win32Resources

ds5678:
I was able to add trimming for:
* `AsmResolver`
* `AsmResolver.PE`
* `AsmResolver.PE.File`
* `AsmResolver.PE.Win32Resources`

I was not able to add trimming to `AsmResolver.DotNet`. Problems originated from only 3 files:
* `DynamicMethodHelper.cs`
* `ReferenceImporter.cs`
* `RuntimeConfiguration.cs`

Annotating these couple uses of json and reflection quickly devolved into me needing to annotate most of the library.

### DIFF
--- a/src/AsmResolver.DotNet/Code/Cil/CilMethodBody.cs
+++ b/src/AsmResolver.DotNet/Code/Cil/CilMethodBody.cs
@@ -136,6 +136,7 @@ namespace AsmResolver.DotNet.Code.Cil
         /// <param name="method">The method that owns the method body.</param>
         /// <param name="dynamicMethodObj">The Dynamic Method/Delegate/DynamicResolver.</param>
         /// <returns>The method body.</returns>
+        [System.Diagnostics.CodeAnalysis.RequiresUnreferencedCode("Calls ResolveDynamicResolver")]
         public static CilMethodBody FromDynamicMethod(MethodDefinition method, object dynamicMethodObj)
         {
             if (!(method.Module is SerializedModuleDefinition module))
--- a/src/AsmResolver.DotNet/Config/Json/RuntimeConfiguration.cs
+++ b/src/AsmResolver.DotNet/Config/Json/RuntimeConfiguration.cs
@@ -1,6 +1,5 @@
 using System.IO;
 using System.Text.Json;
-using System.Text.Json.Serialization;
 
 namespace AsmResolver.DotNet.Config.Json
 {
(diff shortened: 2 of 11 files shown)


Washi1337:
I am not too familiar with the trimming features of .NET 6, so I'll have to look into it. But maybe it would be an option to provide separate build configurations that would exclude the reflection parts that exist in `AsmResolver.DotNet`, since they are only used for very specific use-cases. As for the RuntimeConfiguration part, we could potentially leverage source gen for this, if such a thing for JSON parsers exists. I have not been a huge fan of the `System.Text.Json` dependency anyways, even though it is a great library.

SamboyCoding:
System.Text.Json does support source generation, though I don't know whether the generated source goes "deep enough" to remove all reflection uses.

ds5678:
@Washi1337 can this be part of `4.10.0`? All my changes thus far have been non-breaking and allow trimming for 4 of the packages.

My recent two commits address the issues in `DynamicMethodHelper.cs` and `RuntimeConfiguration.cs`. As a side effect of this, serializing `RuntimeConfiguration` should be faster now that source generation is used.

Any changes regarding `RuntimeConfiguration.cs` would be breaking and deserve a separate pull request for `5.0.0`.

Washi1337:
@ds5678 Thanks for looking into it. A quick glance tells me that indeed this can be put in 4.10.0. I will try to review everything this week (including the release of 4.10.0).  I just have been particularly busy the past couple of days.

Washi1337:
I quickly looked into the warnings that get emitted by the compiler when `IsTrimmable` is added to `AsmResolver.DotNet`. As far as I can tell, the "only" issues that is reported are the calls to `ResolveMethod` and `ResolveField`. However, since the metadata token is pulled from objects from the same module, I think this _should_ be fine right? I know admittedly too little about the new trimming feature, is there some easy way to test this that you know of? Because if it works fine, then I feel like we can just suppress this warning. and enable trimming on this package as well.

Washi1337:
Checking this in more depth reveals that trimming complicates more things than just some token resolution in `ReferenceImporter`.  For example, `TypeSignature` also uses reflection to deal with internal type signatures stored in dynamic methods. 

I think it is best if we leave out trimming in `AsmResolver.DotNet` for now, and make it a separate issue that we can have a look at in more detail in the future. I will merge this as it is now in 4.10.0, so that we can get that pushed on nuget at least.

PR #322, opened 2022-05-29
Fix ExpandMacro.

N78750469:
The expanded instructions of Ldarga_S and Bge_Un_S are wrong.


Washi1337:
Thanks a lot for spotting, and apologies for the late response.

Issue #333, opened 2022-07-16
Callbacks implementation for MemberCloner


LambdaTheDev:
### Problem Description

When working with MemberCloner, you must do at least 2 loops through all copied types - implicit one (in MemberCloner), and explicit one (when you add types to module).

For small assemblies it won't matter, but for larger - it could cause huge processing time.

Callbacks system can solve this - user can define their callback types, cache processed types (for example interfaces), do some minor processing, and add result automatically to module - in a single iteration.

### Proposal

To MemberCloner constructor, add argument `Action<IMemberDescriptor>`, and invoke it directly after the member is fully cloned.

Washi1337:
I can definitely see value in this.

Perhaps as an alternative to calbacks we could also consider implementing a listener pattern for easier customization and case distinction from the user's perspective. This is particularly useful for users that just want to add their included types to the target module, and thus only need a case for `TypeDefinition`. Something like the following:

```csharp
public class MyListener : MemberClonerListener
{
   public override void OnClonedMember(IMetadataMember original, IMetadataMember cloned)
   {
      // called for every member.
   }

   public override void OnClonedType(TypeDefinition original, TypeDefinition cloned)
   {
      // called for every type only
   }

   public override void OnClonedMethod(MethodDefinition original, MethodDefinition cloned)
   {
      // called for every method only.
   }

   // ...
}


var cloner = new MemberCloner(targetModule, new MyListener());
```

We can easily do both as well. The callback version is just a special case of a full listener version, as can be seen in a potential implementation below:

```csharp
public class CallbackCloneListener : MemberClonerListener
{
   private readonly Action<IMetadataMember, IMetadataMember> _callback;

   public CallbackCloneListener(Action<IMetadataMember, IMetadataMember> callback)
   {
      _callback = callback;
   }

   public override void OnClonedMember(IMetadataMember original, IMetadataMember cloned) => _callback(original, cloned);
}

var cloner = new MemberCloner(targetModule, (original, cloned) => { ... }); 
// this would be shorthand for;
// var cloner = new MemberCloner(targetModule, new CallbackCloneListener((original, cloned) => { ... }));
```

LambdaTheDev:
Oh, MemberClonerListener idea is even better (split on types, methods, etc)!

Washi1337:
Implemented in #337. This will be part of the 5.0 release.

PR #337, opened 2022-08-06
Callbacks implementation for MemberCloner

CursedLand:
Solves [#333]

### DIFF
--- a/src/AsmResolver.DotNet/Cloning/CallbackCloneListener.cs
+++ b/src/AsmResolver.DotNet/Cloning/CallbackCloneListener.cs
@@ -2,8 +2,10 @@ using System;
 
 namespace AsmResolver.DotNet.Cloning
 {
-    /// <inheritdoc/>
-    public class CallbackCloneListener : IMemberClonerListener
+    /// <summary>
+    /// This implementation that calls the <see cref="MemberClonerListener.OnClonedMember"/> to a callback action.
+    /// </summary>
+    public class CallbackCloneListener : MemberClonerListener
     {
 
         private readonly Action<IMetadataMember, IMetadataMember> _callback;
@@ -16,7 +18,7 @@ namespace AsmResolver.DotNet.Cloning
             _callback = callback;
 
         /// <inheritdoc/>
-        public void OnClonedMember(IMetadataMember original, IMetadataMember cloned) =>
+        public override void OnClonedMember(IMetadataMember original, IMetadataMember cloned) =>
             _callback(original, cloned);
     }
 }
--- a/src/AsmResolver.DotNet/Cloning/CallbackCloneListener.cs
+++ b/src/AsmResolver.DotNet/Cloning/CallbackCloneListener.cs
@@ -1,0 +1,22 @@
+using System;
+
+namespace AsmResolver.DotNet.Cloning
+{
+    /// <inheritdoc/>
+    public class CallbackCloneListener : IMemberClonerListener
+    {
+
+        private readonly Action<IMetadataMember, IMetadataMember> _callback;
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="CallbackCloneListener"/> class.
+        /// </summary>
+        /// <param name="callback">The Callback used.</param>
+        public CallbackCloneListener(Action<IMetadataMember, IMetadataMember> callback) =>
+            _callback = callback;
+
+        /// <inheritdoc/>
+        public void OnClonedMember(IMetadataMember original, IMetadataMember cloned) =>
+            _callback(original, cloned);
+    }
+}
--- a/src/AsmResolver.DotNet/Cloning/IMemberClonerListener.cs
+++ b/src/AsmResolver.DotNet/Cloning/IMemberClonerListener.cs
@@ -1,15 +1,45 @@
 namespace AsmResolver.DotNet.Cloning
 {
     /// <summary>
-    /// <see cref="MemberCloner"/> Callback Listener.
+    /// <see cref="MemberCloner"/> Callback listener that receives calls after cloning process.
     /// </summary>
     public interface IMemberClonerListener
     {
         /// <summary>
-        /// This function is called for every member got cloned.
+        /// This function is called for every member that got cloned.
         /// </summary>
-        /// <param name="original">original member.</param>
-        /// <param name="cloned">cloned member.</param>
+        /// <param name="original">The original member.</param>
+        /// <param name="cloned">The cloned member.</param>
         public void OnClonedMember(IMetadataMember original, IMetadataMember cloned);
+        /// <summary>
+        /// This function is called for every type got cloned.
+        /// </summary>
+        /// <param name="original">The original type.</param>
+        /// <param name="cloned">The cloned type.</param>
+        public void OnClonedType(TypeDefinition original, TypeDefinition cloned);
+        /// <summary>
+        /// This function is called for every method got cloned.
+        /// </summary>
+        /// <param name="original">The original method.</param>
+        /// <param name="cloned">The cloned method.</param>
+        public void OnClonedMethod(MethodDefinition original, MethodDefinition cloned);
+        /// <summary>
+        /// This function is called for every field got cloned.
+        /// </summary>
+        /// <param name="original">The original field.</param>
+        /// <param name="cloned">The cloned field.</param>
+        public void OnClonedField(FieldDefinition original, FieldDefinition cloned);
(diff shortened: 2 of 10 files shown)


src/AsmResolver.DotNet/Cloning/CallbackCloneListener.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+
+namespace AsmResolver.DotNet.Cloning
+{
+    /// <inheritdoc/>
```
Washi1337, line 5: We probably want to add some docs on what the specific purpose is of this implementation of the `IMemberClonerListener`. i.e., explain that this implementation that maps the `OnClonedMember` to a delegate.

src/AsmResolver.DotNet/Cloning/IMemberClonerListener.cs
```diff
@@ -0,0 +1,15 @@
+namespace AsmResolver.DotNet.Cloning
+{
+    /// <summary>
+    /// <see cref="MemberCloner"/> Callback Listener.
```
Washi1337, line 4: Similar as above, we need to explain what the purpose is of this interface. i.e., that this interface's members receive notifications from the cloner.

src/AsmResolver.DotNet/Cloning/IMemberClonerListener.cs
```diff
@@ -0,0 +1,15 @@
...
+    /// </summary>
+    public interface IMemberClonerListener
+    {
+        /// <summary>
+        /// This function is called for every member got cloned.
```
Washi1337, line 9: This function is called for every member **that** got cloned.

src/AsmResolver.DotNet/Cloning/IMemberClonerListener.cs
```diff
@@ -0,0 +1,15 @@
...
+        /// </summary>
+        /// <param name="original">original member.</param>
```
Washi1337, line 11: Use capitalization. i.e., `The original member`

src/AsmResolver.DotNet/Cloning/MemberCloner.cs
```diff
@@ -30,18 +31,29 @@ public partial class MemberCloner
         /// Creates a new instance of the <see cref="MemberCloner"/> class.
         /// </summary>
         /// <param name="targetModule">The target module to copy the members into.</param>
-        public MemberCloner(ModuleDefinition targetModule) : this(targetModule, null) { }
+        /// <param name="callback">The callback used in the cloner listener.</param>
+        public MemberCloner(ModuleDefinition targetModule, Action<IMetadataMember, IMetadataMember>? callback = null) : this(targetModule, null, new CallbackCloneListener(callback ?? new Action<IMetadataMember, IMetadataMember>((orginal, cloned) => { }))) { }
```
Washi1337, line 35: Please break up long lines :)

Washi1337: Also, adding the optional parameter is a breaking ABI change, which means it has to wait till 5.0. If we want this to get into 4.x still, we should maybe split up this constructor into two overloaded instead.

src/AsmResolver.DotNet/Cloning/MemberCloner.Fields.cs
```diff
@@ -39,7 +39,13 @@ private static FieldDefinition CreateFieldStub(MemberCloneContext context, Field
         private void DeepCopyFields(MemberCloneContext context)
         {
             foreach (var field in _fieldsToClone)
+            {
                 DeepCopyField(context, field);
+                var clonedMember = (FieldDefinition)context.ClonedMembers[field];
+                _clonerListener.OnClonedMember(field, clonedMember);
+                if (_clonerListener is MemberClonerListener listener)
```
Washi1337, line 46: I think we should move this pattern matching (and all similar subsequent matchings) out of the `MemberCloner` class and into `MemberClonerListener::OnClonedMember` (i.e. with a switch case). Hardwiring to a specific implementation of an interface is kind of an anti-pattern. It restricts how the cloner and the listener can be used.

How about we move all `OnClonedXXX` from `MemberClonerListener` into the `IMemberClonerInterface` interface directly, and remove the generic `OnClonedMember` from the interface, but keep it in the abstract class? We can make then everything `virtual` in `MemberClonerListener` and let the default implementations just call `OnClonedMember`. We can then also let `CallbackClonerListener` inherit from the base class instead of the interface. Should get rid of all these `if`s in `MemberCloner`, and also makes it easier to make your own listeners. What do you think?

src/AsmResolver.DotNet/Cloning/MemberClonerListener.cs
```diff
@@ -0,0 +1,41 @@
+namespace AsmResolver.DotNet.Cloning
+{
+    /// <summary>
+    /// <see cref="MemberCloner"/> Callback Listener.
+    /// </summary>
+    public abstract class MemberClonerListener : IMemberClonerListener
+    {
+        /// <inheritdoc/>
+        public abstract void OnClonedMember(IMetadataMember original, IMetadataMember cloned);
```
Washi1337, line 9: I think it makes sense to make these `virtual` rather than `abstract`. This makes it easier for consumers to use the base class, as they only have to override the methods they are interested in overriding. With everything abstract, consumers gain nothing over from just implementing the interface.

test/AsmResolver.Tests/Listeners/CustomMemberClonerListener.cs
```diff
@@ -0,0 +1,26 @@
+using AsmResolver.DotNet;
+using AsmResolver.DotNet.Cloning;
+using System;
+
+namespace AsmResolver.Tests.Listeners {
+    public class CustomMemberClonerListener : MemberClonerListener {
```
Washi1337, line 6: Brace on new-line.

test/AsmResolver.DotNet.Tests/Cloning/MetadataClonerTest.cs
```diff
@@ -329,5 +330,52 @@ public void CloneInterfaceImplementations()
...
+            var clonedType = result.GetClonedMember(type);
+
+            Assert.Equal(
+                type.Methods.Select(m => m.Name.Reverse().ToArray()),
+                clonedType.Methods.Select(m => m.Name.ToArray()));
+
+        }
+
+        [Fact]
+        public void CloneCustomListenerResult() {
```
Washi1337, line 362: Brace on new line

test/AsmResolver.Tests/Listeners/CustomMemberClonerListener.cs
```diff
@@ -0,0 +1,26 @@
...
+        public override void OnClonedEvent(EventDefinition original, EventDefinition cloned) =>
+            cloned.Name = $"Event_{original.Name}";
+
+        public override void OnClonedField(FieldDefinition original, FieldDefinition cloned) =>
+            cloned.Name = $"Field_{original.Name}";
+
+        public override void OnClonedMember(IMetadataMember original, IMetadataMember cloned) {
```
Washi1337, line 13: Both braces on new-line, or both on same line.

Washi1337 (changes requested):
Nice, thanks a lot for writing an implementation and tests!

Few things here and there regarding code style inconsistencies (maybe check `.editorconfig`), but I think the main thing that we probably want to consider changing is making all methods `virtual` instead of `abstract`, and moving the `OnClonedXXX` members into the interface. See comments below.

Other than that, lgtm.

src/AsmResolver.DotNet/Cloning/IMemberClonerListener.cs
Washi1337: You probably want to fix the same grammar mistake for every member, not just OnClonedMember :p

src/AsmResolver.DotNet/Cloning/MemberClonerListener.cs
```diff
@@ -0,0 +1,18 @@
+namespace AsmResolver.DotNet.Cloning
+{
+    /// <inheritdoc/>
+    public abstract class MemberClonerListener : IMemberClonerListener {
```
Washi1337, line 4: Brace on new line, you might want to setup `.editorconfig` so that these code style layout inconsistencies don't happen :).

CursedLand: oh VS betrayed me again xd

PR #338, opened 2022-08-07
Make TypeDefOrRefSignature.Type settable

JPaja:


Washi1337:
I think if we do this, including the automated `IsValueType` update, then we should also do this for `GenericInstanceTypeSignature` for consistency. If you could do that then I will merge this :)

Washi1337:
Thanks a lot!

Washi1337 (approved):

PR #360, opened 2022-09-30
Improve Leaf Record Errors

ds5678:


src/AsmResolver.Symbols.Pdb/Records/CodeViewSymbol.cs
```diff
@@ -37,4 +38,19 @@ public static CodeViewSymbol FromReader(PdbReaderContext context, ref BinaryStre
...
+    private protected T? GetLeafRecord<T>(PdbReaderContext context, uint typeIndex) where T : CodeViewLeaf
+    {
+        return context.ParentImage.TryGetLeafRecord(typeIndex, out var leaf)
+            ? leaf switch
+            {
+                T t => t,
+                UnknownCodeViewLeaf unknownLeaf => context.Parameters.ErrorListener.BadImageAndReturn<T>(
+                    $"{GetType().Name} contains an unknown leaf ({unknownLeaf.LeafKind}) at index {typeIndex:X8}."),
+                _ => context.Parameters.ErrorListener.BadImageAndReturn<T>(
+                    $"{GetType().Name} contains an incorrect leaf type ({leaf.GetType().Name}) at index {typeIndex:X8}.")
```
Washi1337, line 51: To me the error message is a bit confusing now, as it is unclear what `typeIndex` is referring to now (the referenced leaf or the index of the leaf that is referencing the other leaf?).

Would it make more sense to rephrase this error as `"{GetType().Name} {TypeIndex:X8} references a leaf at {typeIndex:X8} that is of an unexpected type ({leaf.GetType().Name})."`?

Washi1337 (changes requested):
I like the idea, but I think if we want to capture the unknown / unsupported type records, then we should be consistent with this and use it everywhere. E.g., [SerializedArrayTypeRecord](https://github.com/Washi1337/AsmResolver/blob/e4ed4f09f13da9b3db9a5bd4f9888db7020fca60/src/AsmResolver.Symbols.Pdb/Leaves/Serialized/SerializedArrayTypeRecord.cs#L36-L42) could benefit from the same type of construction.

However, simply replacing code like this with a call to the proposed `GetLeafRecord` method loses contextual information if an error occurs. E.g., there is a difference between a message such as `SerializedArrayTypeRecord contains an incorect leaf type at index 01234.` and `Array type 5678 contains an invalid element type index 1234.`. The second one is more specific, since it also indicates which property failed to be parsed,  rather than simply stating that _some_ index within the leaf was invalid. Furthermore, the first does not include the type index of the faulty leaf itself either.

Can we somehow rephrase / refactor this such that we do not lose this contextual information?

src/AsmResolver.Symbols.Pdb/Records/CodeViewSymbol.cs
Washi1337: Also, we could probably avoid the `GetType().Name` and replace it with `LeafKind` instead.

ds5678: I standardized the wording based on your suggestion.

ds5678:
How do you want to handle the `SerializedArrayTypeRecord` situation?

Washi1337:
@ds5678 The array type record was just an example of how a leaf could reference multiple type indices, and thus it would be unclear in the error message of the proposed change _which_ type index was invalid. 

Some simple idea would be to add a string parameter to the `GetLeafRecord` method, indicating which type index was attempted to be looked up, so it could be used in the error string format. I am open for other ideas as well though.

Washi1337:
Closed in favor of other previously merged PRs

PR #367, opened 2022-10-16
Use IErrorListener while building managed assemblies

ds5678:


src/AsmResolver.DotNet/Builder/PEImageBuildContext.cs
```diff
@@ -12,24 +12,43 @@ public class PEImageBuildContext
...
+
+        /// <summary>
+        /// Gets the bag that collects all diagnostic information during the building process. 
+        /// </summary>
+        [Obsolete]
         public DiagnosticBag DiagnosticBag
+        {
+            get => (DiagnosticBag)ErrorListener;
+        }
+
```
Washi1337, line 45: We may want to define this as a `DiagnosticBag?`, and use `as` instead of a direct cast to avoid casting exceptions. Also, let's notify consumers using the `Obsolete` CA that the `ErrorListener` is now preferred over the `DiagnosticBag`.

src/AsmResolver.DotNet/Builder/PEImageBuildContext.cs
```diff
@@ -12,24 +12,43 @@ public class PEImageBuildContext
...
-        /// <param name="diagnosticBag">The diagnostic bag to use.</param>
+        /// <param name="diagnosticBag">The diagnostic bag to use.</param>
+        [Obsolete]
         public PEImageBuildContext(DiagnosticBag diagnosticBag)
         {
-            DiagnosticBag = diagnosticBag ?? throw new ArgumentNullException(nameof(diagnosticBag));
-        }
-        
+            ErrorListener = diagnosticBag ?? throw new ArgumentNullException(nameof(diagnosticBag));
+        }
```
Washi1337, line 26: Maybe mention in the `Obsolete` attribute what the new preferred constructor is.

src/AsmResolver.DotNet/ModuleDefinition.cs
```diff
@@ -1295,14 +1295,18 @@ public void Write(IBinaryStreamWriter writer, IPEImageBuilder imageBuilder, IPEF
...
+            if (result.ErrorListener is DiagnosticBag diagnosticBag && diagnosticBag.HasErrors)
             {
                 throw new AggregateException(
                     "Construction of the PE image failed with one or more errors.",
-                    result.DiagnosticBag.Exceptions);
+                    diagnosticBag.Exceptions);
+            }
+            else if (result.HasFailed)
+            {
+                throw new Exception("Construction of the PE image failed.");
```
Washi1337, line 1306: Let's not use the generic `Exception`, but maybe use `MetadataBuilderException` instead, or else define a new exception type (maybe `PEImageBuilderException`?).

Washi1337 (changes requested):
I think this is a good compromise until we make full breaking changes in 6.0. I have only small points of improvement, but otherwise LGTM.

PR #368, opened 2022-10-16
Add support for reading LProcRef and ProcRef symbols

ds5678:


Washi1337:
Would be nice if there are some unit test, to ensure that we don't break the implementation in the future. The `SimplePdb` resource has some `S_PROCREF` and  `S_LPROCREF` entries that we can verify.

Other than that, LGTM!

Washi1337 (approved):
LGTM. Thanks!

PR #369, opened 2022-10-20
BUGFIX: resolving m_scope field from DynamicILInfo rather than DynamicResolver.

CursedLand:


Washi1337 (approved):
Thanks a lot for implementing the fix!

PR #390, opened 2022-12-18
Fix docs order of instructions

sunnamed434:
[#Referencing members](https://asmresolver.readthedocs.io/en/development/dotnet/managed-method-bodies.html#referencing-members) has wrong instructions order (Console.WriteLine() without Ldstr when in Type[] it asks for string argument)


Washi1337:
Thanks!

PR #391, opened 2022-12-20
Add new property that whether returns a value that indicates if the T…

sunnamed434:
…ypeDefinition is `<Module>` type


sunnamed434:
Idk how you sort the usings, I just pressed the hotkey CTRL+R+G

sunnamed434:
To me this seems much more better

Washi1337:
Thanks!

PR #418, opened 2023-02-28
Add Parameter.CreateParameterDefinition helper method

ElektroKill:
Changes:
* Added a new method to `Parameter` to make the creation of `ParameterDefinition` objects easier
* Don't use unbound generic types in virtual hidden this parameter.
* Fixed a bug that resulted in incorrect parameter count calculation for signatures with `ExplicitThis` flag set to true.

### DIFF
--- a/src/AsmResolver.DotNet/Collections/Parameter.cs
+++ b/src/AsmResolver.DotNet/Collections/Parameter.cs
@@ -76,6 +76,11 @@ namespace AsmResolver.DotNet.Collections
         /// <inheritdoc />
         public string Name => Definition?.Name ?? GetDummyArgumentName(MethodSignatureIndex);
 
+        /// <summary>
+        /// Creates a <see cref="ParameterDefinition"/> for this parameter and adds it to the method definition if it does not exist.
+        /// </summary>
+        public void CreateParameterDefinition() => _parentCollection?.CreateParameterDefinition(this);
+
         [SuppressMessage("ReSharper", "InconsistentlySynchronizedField")]
         private static string GetDummyArgumentName(int index)
         {
--- a/src/AsmResolver.DotNet/Collections/Parameter.cs
+++ b/src/AsmResolver.DotNet/Collections/Parameter.cs
@@ -80,8 +80,12 @@ namespace AsmResolver.DotNet.Collections
         /// Creates a or returns the existing <see cref="ParameterDefinition"/> corresponding to this parameter.
         /// If a <see cref="ParameterDefinition"/> is created it is automatically added to the method definition.
         /// </summary>
-        /// <returns>The created or existing <see cref="ParameterDefinition"/> or null if this parameter is virtual.</returns>
-        public ParameterDefinition? GetOrCreateDefinition() => _parentCollection?.GetOrCreateParameterDefinition(this);
+        public ParameterDefinition GetOrCreateDefinition()
+        {
+            if (_parentCollection is null)
+                throw new InvalidOperationException("Cannot create a parameter definition for a parameter that has been removed from its parent collection.");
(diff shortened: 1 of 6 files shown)


src/AsmResolver.DotNet/Collections/Parameter.cs
```diff
@@ -76,6 +76,11 @@ public TypeSignature ParameterType
         /// <inheritdoc />
         public string Name => Definition?.Name ?? GetDummyArgumentName(MethodSignatureIndex);
 
+        /// <summary>
+        /// Creates a <see cref="ParameterDefinition"/> for this parameter and adds it to the method definition if it does not exist.
+        /// </summary>
+        public void CreateParameterDefinition() => _parentCollection?.CreateParameterDefinition(this);
```
Washi1337, line 82: Should we rename it to `CreateDefinition` for brevity and let it return the created `ParameterDefinition`? It is already clear from the context that this definition is going to be a parameter definition, and it would allow for immediately using the created definition as a variable as well.

src/AsmResolver.DotNet/Collections/ParameterCollection.cs
```diff
@@ -142,6 +155,13 @@ private void UpdateParameterTypes()
             return _owner.ParameterDefinitions.FirstOrDefault(p => p.Sequence == sequence);
         }
 
+        internal void CreateParameterDefinition(Parameter parameter)
+        {
+            if (parameter == ThisParameter || parameter.Definition is not null)
+                return;
+            _owner.ParameterDefinitions.Add(new ParameterDefinition(parameter.Sequence, Utf8String.Empty, 0));
+        }
+
```
Washi1337, line 164: The name `CreateParameterDefinition` suggests that it always creates a new parameter, however the implementation doesn't do this when the definition already exists. Should we throw instead? We could perhaps also introduce a `GetOrCreateDefinition` instead.

test/AsmResolver.DotNet.Tests/Collections/ParameterCollectionTest.cs
```diff
@@ -111,6 +118,18 @@ public void ReadMultipleParametersFromInstanceMethod()
...
         }
 
+        [Fact]
+        public void ReadEmptyParametersFromGenericInstanceMethod()
+        {
+            var method = ObtainGenericInstanceTestMethod(nameof(GenericInstanceMethods<int, int>.InstanceParameterlessMethod));
+            Assert.Empty(method.Parameters);
+            Assert.NotNull(method.Parameters.ThisParameter);
+            var genericInstanceType = method.Parameters.ThisParameter.ParameterType as GenericInstanceTypeSignature;
+            Assert.NotNull(genericInstanceType);
```
Washi1337, line 128: Replace with 
```csharp
var genericInstanceType = Assert.IsAssignableFrom<GenericInstanceTypeSignature>(method.Parameters.ThisParameter.ParameterType);
```

test/AsmResolver.DotNet.Tests/Collections/ParameterCollectionTest.cs
```diff
@@ -111,6 +118,18 @@ public void ReadMultipleParametersFromInstanceMethod()
...
+            Assert.True(genericInstanceType.TypeArguments.Count == 2);
+            Assert.Equal("GenericInstanceMethods`2", genericInstanceType.GenericType.Name);
```
Washi1337, line 130: Should we also test for the type arguments all being the right generic type parameter signatures pointing to the right indices?

Washi1337 (commented):
Some minor comments, but otherwise LGTM.

Washi1337 (approved):
Thanks for implementing this!

PR #483, opened 2023-09-18
Add basic Span<T> support

DaZombieKiller:
This pull request adds basic support for `Span<T>` and `ReadOnlySpan<T>` to the `AsmResolver` and `AsmResolver.IO` namespaces.

Support is only implemented for .NET Standard 2.1 and higher, no attempt has been made to provide `Span<T>` support for .NET Standard 2.0 as it would introduce a dependency on `System.Memory`.

This PR also does not attempt to take advantage of this new support within the library. This is most likely better left to a separate PR (and/or set of issues).

The following new types are provided to avoid source and binary breaking changes in existing code:
```cs
namespace AsmResolver.IO
{
    public interface ISpanDataSource : IDataSource
    {
        int ReadBytes(ulong address, Span<byte> buffer);
    }

    public interface ISpanBinaryStreamWriter : IBinaryStreamWriter
    {
        void WriteBytes(ReadOnlySpan<byte> buffer);
    }

    public static class DataSourceExtensions
    {
        public static int ReadBytes(this IDataSource dataSource, ulong address, Span<byte> buffer);
    }
}
```
`ISpanDataSource` and `ISpanBinaryStreamWriter` can be considered temporary, as they only exist to avoid the breaking change of adding a new member to the corresponding `IDataSource` and `IBinaryStreamWriter` interfaces. In the future, it may be desirable to remove them and take the break.

### DIFF
--- a/src/AsmResolver/AsmResolver.csproj
+++ b/src/AsmResolver/AsmResolver.csproj
@@ -7,7 +7,7 @@
         <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
         <NoWarn>1701;1702;NU5105</NoWarn>
         <Nullable>enable</Nullable>
-        <TargetFrameworks>net6.0;netcoreapp3.1;netstandard2.0</TargetFrameworks>
+        <TargetFrameworks>net6.0;netcoreapp3.1;netstandard2.0;netstandard2.1</TargetFrameworks>
         <IsTrimmable>true</IsTrimmable>
     </PropertyGroup>
 
--- a/src/AsmResolver/IO/DataSourceExtensions.cs
+++ b/src/AsmResolver/IO/DataSourceExtensions.cs
@@ -1,0 +1,34 @@
+using System;
+#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP1_0_OR_GREATER
+using System.Buffers;
+#endif
+
+namespace AsmResolver.IO
+{
+    /// <summary>
+    /// Provides extension methods for <see cref="IDataSource"/>.
+    /// </summary>
+    public static class DataSourceExtensions
+    {
+#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER
+        /// <summary>
+        /// Reads a block of data from the data source.
+        /// </summary>
+        /// <param name="dataSource">The <see cref="IDataSource"/> to read from.</param>
+        /// <param name="address">The starting address to read from.</param>
+        /// <param name="buffer">The buffer that receives the read bytes.</param>
+        /// <returns>The number of bytes that were read.</returns>
+        public static int ReadBytes(this IDataSource dataSource, ulong address, Span<byte> buffer)
+        {
+            if (dataSource is ISpanDataSource spanDataSource)
+                return spanDataSource.ReadBytes(address, buffer);
+
+            byte[] array = ArrayPool<byte>.Shared.Rent(buffer.Length);
+            int bytesRead = dataSource.ReadBytes(address, array, 0, buffer.Length);
+            new ReadOnlySpan<byte>(array, 0, bytesRead).CopyTo(buffer);
+            ArrayPool<byte>.Shared.Return(array);
+            return bytesRead;
+        }
+#endif
+    }
+}
--- a/src/AsmResolver/IO/IBinaryStreamWriter.cs
+++ b/src/AsmResolver/IO/IBinaryStreamWriter.cs
@@ -1,5 +1,8 @@
 using System;
 using System.Text;
+#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP1_0_OR_GREATER
+using System.Buffers;
+#endif
 
 namespace AsmResolver.IO
 {
@@ -129,6 +132,27 @@ namespace AsmResolver.IO
         {
             writer.WriteBytes(buffer, 0, buffer.Length);
         }
+
+#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER
+        /// <summary>
+        /// Writes a buffer of data to the stream.
+        /// </summary>
+        /// <param name="writer">The writer to use.</param>
+        /// <param name="buffer">The data to write.</param>
+        public static void WriteBytes(this IBinaryStreamWriter writer, ReadOnlySpan<byte> buffer)
+        {
+            if (writer is ISpanBinaryStreamWriter spanWriter)
+            {
+                spanWriter.WriteBytes(buffer);
+                return;
+            }
+
+            byte[] array = ArrayPool<byte>.Shared.Rent(buffer.Length);
+            buffer.CopyTo(array);
+            writer.WriteBytes(array, 0, buffer.Length);
+            ArrayPool<byte>.Shared.Return(array);
+        }
+#endif
 
         /// <summary>
         /// Writes a specified amount of zero bytes to the stream.
--- a/src/AsmResolver/IO/MemoryMappedDataSource.cs
+++ b/src/AsmResolver/IO/MemoryMappedDataSource.cs
@@ -7,6 +7,9 @@ namespace AsmResolver.IO
     /// Represents a data source that obtains its data from a memory mapped file.
     /// </summary>
     public sealed class MemoryMappedDataSource : IDataSource, IDisposable
+#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER
+        , ISpanDataSource
+#endif
     {
         private readonly MemoryMappedViewAccessor _accessor;
 
@@ -39,6 +42,24 @@ namespace AsmResolver.IO
         /// <inheritdoc />
         public int ReadBytes(ulong address, byte[] buffer, int index, int count) =>
             _accessor.ReadArray((long) address, buffer, index, count);
+
+#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER
+        /// <inheritdoc />
+        public unsafe int ReadBytes(ulong address, Span<byte> buffer)
+        {
+            var handle = _accessor.SafeMemoryMappedViewHandle;
+            int actualLength = (int) Math.Min(handle.ByteLength, (uint) buffer.Length);
+#if NET6_0_OR_GREATER
+            handle.ReadSpan(address, buffer);
+#else
+            byte* pointer = null;
+            handle.AcquirePointer(ref pointer);
+            new ReadOnlySpan<byte>(pointer, actualLength).CopyTo(buffer);
+            handle.ReleasePointer();
+#endif
+            return actualLength;
+        }
+#endif
 
         /// <inheritdoc />
         public void Dispose() => _accessor?.Dispose();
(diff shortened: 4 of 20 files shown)


ds5678:
It might be a good idea to add a define variable, eg SPAN_SUPPORTED.

src/AsmResolver/IO/IBinaryStreamWriter.cs
```diff
@@ -130,6 +133,27 @@ public static void WriteBytes(this IBinaryStreamWriter writer, byte[] buffer)
...
+            if (writer is ISpanBinaryStreamWriter spanWriter)
+            {
+                spanWriter.WriteBytes(buffer);
+                return;
+            }
+
+            byte[] array = ArrayPool<byte>.Shared.Rent(buffer.Length);
+            buffer.CopyTo(array);
+            writer.WriteBytes(array, 0, buffer.Length);
+            ArrayPool<byte>.Shared.Return(array);
```
Washi1337, line 153: We should probably wrap this into a `try-finally` construction to ensure the rented array is returned even in the event `CopyTo` or `WriteBytes` fail.

src/AsmResolver/IO/MemoryMappedDataSource.cs
```diff
@@ -40,6 +43,24 @@ public ulong Length
...
+        {
+            var handle = _accessor.SafeMemoryMappedViewHandle;
+            int actualLength = (int) Math.Min(handle.ByteLength, (uint) buffer.Length);
+#if NET6_0_OR_GREATER
+            handle.ReadSpan(address, buffer);
+#else
+            byte* pointer = null;
+            handle.AcquirePointer(ref pointer);
+            new ReadOnlySpan<byte>(pointer, actualLength).CopyTo(buffer);
+            handle.ReleasePointer();
```
Washi1337, line 58: Similar comment on try-finally construction here as well.

src/AsmResolver/AsmResolver.csproj
```diff
@@ -7,7 +7,7 @@
         <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
         <NoWarn>1701;1702;NU5105</NoWarn>
         <Nullable>enable</Nullable>
-        <TargetFrameworks>net6.0;netcoreapp3.1;netstandard2.0</TargetFrameworks>
+        <TargetFrameworks>net6.0;netcoreapp3.1;netstandard2.0;netstandard2.1</TargetFrameworks>
```
Washi1337, line 10: Should we add netstandard2.1 targets to the other csproj as well, for consistency?

src/AsmResolver/IO/DataSourceExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <returns>The number of bytes that were read.</returns>
+        public static int ReadBytes(this IDataSource dataSource, ulong address, Span<byte> buffer)
+        {
+            if (dataSource is ISpanDataSource spanDataSource)
+                return spanDataSource.ReadBytes(address, buffer);
+
+            byte[] array = ArrayPool<byte>.Shared.Rent(buffer.Length);
+            int bytesRead = dataSource.ReadBytes(address, array, 0, buffer.Length);
+            new ReadOnlySpan<byte>(array, 0, bytesRead).CopyTo(buffer);
+            ArrayPool<byte>.Shared.Return(array);
```
Washi1337, line 29: Here, `try-finally` as well.

src/AsmResolver/IO/IBinaryStreamWriter.cs
DaZombieKiller: `try`/`finally` is actually discouraged for `ArrayPool` because the program state is potentially undefined in the presence of an exception, so something may still be referencing the rented array as part of its corrupted state, resulting in potentially dangerous behavior later on (there have been some discussions in the C# Discord about making this clearer in the documentation).

Failure to call `ArrayPool<T>.Return` is also considered safe, it just means that the GC will reclaim the array and the operation effectively becomes equivalent to `new T[]` instead of renting.

So a `try`/`finally` here would mostly just serve to slow down the method by adding an exception handler.

src/AsmResolver/AsmResolver.csproj
DaZombieKiller: I considered it, though since they aren't actually making use of these APIs yet I wasn't sure if it would make sense to. Do you have a preference here?

src/AsmResolver/IO/DataSourceExtensions.cs
DaZombieKiller: Same scenario as the other `ArrayPool` usage.

src/AsmResolver/IO/IBinaryStreamWriter.cs
ds5678: Interesting. Today I learned.

DaZombieKiller:
@ds5678
> It might be a good idea to add a define variable, eg `SPAN_SUPPORTED`, instead of using `NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER` for the preprocessor directives.

I thought about doing this, but decided to wait for feedback to see what would be preferred. Since there aren't any plans to expose this under .NET Standard 2.0, `SPAN_SUPPORTED` will never mean anything other than `NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER`, but I suppose it is much less noisy in the source.

src/AsmResolver/IO/IBinaryStreamWriter.cs
Washi1337: Not sure if I fully agree/understand. 

As far as I know, the pool returns instances of arrays while also holding a reference to it. Thus, a clean up of the array only happens if the pool was also cleaned. Thus, the instance would only be reclaimed if the pool was not shared and collected as well. Not returning the array would mean the array would "live forever", even if all objects still holding a reference to this array (other than the pool itself) would be cleaned up.

Additionally, I understand that returning array instances that are potentially still in use due to the exceptional code path that was taken is dangerous. However, in either case, if we assume that the callee (i.e., WriteBytes) does not pass along the array to anywhere else (which I think is a pretty fair assumption given the semantics of this method), the arrays are not used anywhere else anyways, never reaching such an invalid or corrupted state.

Or am I missing something here?

src/AsmResolver/AsmResolver.csproj
Washi1337: Let's add them to all packages, even if it is just for consistency sake.

Washi1337:
I have no real strong opinions on introducing either `SPAN_SUPPORTED` or leaving it as an OR expression as it is now. If it is easy to add `SPAN_SUPPORTED` without too much messing around in the csprojs, then I'm all for it. If not, we can leave it as is as well.

src/AsmResolver/IO/IBinaryStreamWriter.cs
ds5678: https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Buffers/ArrayPool.cs,73

ds5678: If I'm understanding this correctly, it's also not required that the array being returned was rented. It seems valid for a coder to "donate" an array to the pool, though dubious in usefulness and efficiency.

DaZombieKiller: > As far as I know, the pool returns instances of arrays while also holding a reference to it. Thus, a clean up of the array only happens if the pool was also cleaned. Thus, the instance would only be reclaimed if the pool was not shared and collected as well. Not returning the array would mean the array would "live forever", even if all objects still holding a reference to this array (other than the pool itself) would be cleaned up.

[The pool itself doesn't hold a reference to an array after it has been rented](https://github.com/dotnet/runtime/blob/d05415704826f0bd084d2c27f99e4dbe74f62464/src/libraries/System.Private.CoreLib/src/System/Buffers/SharedArrayPool.cs#L66). That is done specifically to allow the GC to reclaim arrays that were not returned to the pool (which is considered a safe thing to do, and the BCL does it in several places to avoid these issues).

> Additionally, I understand that returning array instances that are potentially still in use due to the exceptional code path that was taken is dangerous. However, in either case, if we assume that the callee (i.e., WriteBytes) does not pass along the array to anywhere else (which I think is a pretty fair assumption given the semantics of this method), the arrays are not used anywhere else anyways, never reaching such an invalid or corrupted state.

If you know for certain that the code will not be holding onto the array, then it's fine to have a `try`/`finally` yeah (it just slightly pessimizes the codegen of the method and makes inlining much less likely to occur, if at all). Generally it's only really *desirable* to have the `finally` if that condition is met _and_ you expect there to be a lot of exceptions thrown (in a cancellable async method, for example) so that you don't starve the ArrayPool. Since this is being passed along to user code though, *technically* no guarantee can be made (imagine if that user code innocently passed the array along to a `Stream` that held onto the array in the presence of an exception, for example).

> EDIT: I guess there is an argument that because it is virtual anyone could override the method and make it do something that will keep the reference somewhere... Though I feel bad implementations are always going to be bad, regardless of correct/misuse of the pool.

My comments specifically apply to `ArrayPool<T>.Shared` since that's what's being used here, but if the API took an `ArrayPool<T>` as input then this would be a consideration as well, yeah.

DaZombieKiller: For some further context, [see this comment](https://github.com/dotnet/runtime/issues/71249#issuecomment-1165584805) (and the surrounding thread in general).

src/AsmResolver/IO/ZeroesDataSource.cs
```diff
@@ -54,5 +57,15 @@ public int ReadBytes(ulong address, byte[] buffer, int index, int count)
             Array.Clear(buffer, index, count);
             return actualLength;
         }
+
+#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER
+        /// <inheritdoc />
+        public int ReadBytes(ulong address, Span<byte> buffer)
+        {
+            int actualLength = (int) Math.Min(Length, (uint) buffer.Length);
+            buffer.Clear();
```
DaZombieKiller, line 66: The code here is clearing the entire input buffer, which matches the current behavior of the corresponding `byte[]` overload. However, this seems like a performance trap if a large buffer is passed in, since technically only `[0..actualLength)` needs to be cleared.

Was this intentional behavior? If not, I could change this to `buffer[..actualLength].Clear();` and change the `byte[]` overload to use `Array.Clear(buffer, index, actualLength);` to avoid doing more work in the method than necessary.

Washi1337: This is an oversight. Your proposed change makes sense, also for the original API. If you can change it that would be great!

src/AsmResolver/IO/IBinaryStreamWriter.cs
Washi1337: > [The pool itself doesn't hold a reference to an array after it has been rented](https://github.com/dotnet/runtime/blob/d05415704826f0bd084d2c27f99e4dbe74f62464/src/libraries/System.Private.CoreLib/src/System/Buffers/SharedArrayPool.cs#L66). That is done specifically to allow the GC to reclaim arrays that were not returned to the pool (which is considered a safe thing to do, and the BCL does it in several places to avoid these issues).

Interesting, TIL. This is the core of my misunderstanding. If the pool does not keep a reference to the array after it is rented, this seems fine.

Washi1337 (approved):

Washi1337:
I just realized your PR is based on `master`.  Can you please change the base branch to `development` and see if all the tests still succeed?

Washi1337:
Thanks!

PR #512, opened 2024-01-01
Fix a couple of corner cases in SignatureComparer.GetHashCode

rstarkov:
Just fixing three instances of object.GetHashCode being called by mistake in `GenericInstanceTypeSignature` and `CustomModifierTypeSignature`. I had a few fields and methods in my assemblies where .Equals returned true (correctly) but .GetHashCode returned distinct hash codes.


Washi1337 (approved):

Washi1337:
Good catch, thanks!

PR #523, opened 2024-02-24
Fix function pointer Type imports

Windows10CE:


src/AsmResolver.DotNet/ReferenceImporter.cs
```diff
@@ -332,6 +367,71 @@ public virtual TypeSignature ImportTypeSignature(Type type)
             return new TypeDefOrRefSignature(reference, type.IsValueType);
         }
 
+        // separate class to prevent static cctor checks from leaking into the other ReferenceImporter methods on NAOT
+        private static class FnPtrHelper
+        {
```
Washi1337, line 372: This class seems somewhat weirdly positioned in between all other methods. We may want to move it to a separate file solely related to these types of helpers. We can then also move the `Marshal.GetHINSTANCE` reflection hack in [ModuleDefinition::FromModule](https://github.com/Washi1337/AsmResolver/blob/6f0a38370372ffc8029a5c1409ea6611bdd4c6a8/src/AsmResolver.DotNet/ModuleDefinition.cs#L171). Besides, I also expect we will get more reflection "hacks" like this in the future.

test/AsmResolver.DotNet.Tests/ReferenceImporterTest.cs
```diff
@@ -546,5 +546,79 @@ public void ImportGenericMethodInstantiationWithReturnTypeViaReflection()
             var signature = Assert.IsAssignableFrom<GenericParameterSignature>(imported.Signature.ReturnType);
             Assert.Equal(2, signature.Index);
         }
+
+        private static unsafe class DelegatePointerHolder
+        {
+            public static delegate*unmanaged[Cdecl, SuppressGCTransition]<int, uint> Complex;
+            public static delegate*unmanaged[Stdcall]<int, uint> StdcallOnly;
+            public static delegate*unmanaged[SuppressGCTransition]<int, uint> GcOnly;
+        }
```
Washi1337, line 555: Should we guard this as well with a `#if !NET8_0_OR_GREATER`?

Washi1337 (changes requested):
Few nitpicks but other than that LGTM!

Washi1337 (approved):

Washi1337:
Thanks for looking into it!

PR #540, opened 2024-03-17
.NET 3.5 via MonoMod.Backports

### LINKED ISSUE #536: Add .NET 3.5 Targets
### Problem Description

There has been some demand for .NET 3.5 targets. In particular, use cases include:
- Dynamic analysis / deobfuscation of binaries targeting .NET 3.5
- Game mods that target games running Unity targeting .NET 3.5 Mono.

Currently, AsmResolver's is compatible with .NET standard 2.0 or higher. This means it requires at the very least .NET 4.6.1. We therefore miss out on all usecases that require an older version.

### Proposal

Add a .NET 3.5 build target.

This is not as trivial as it may seem, due to the following incompatibilities that AsmResolver currently depends on:
- [x] `System.Threading.Concurrent`. See #540 
- [x] `System.Text.Json`. See #450 
- [x] `System.ValueTuple`. See #540 
- [x] `System.IO.MemoryMappedFiles`. See #544 

These will have to be polyfilled or shims have to be written.

### Alternatives

As 3.5 is ancient technology, we could opt for not supporting .NET 3.5. However, unfortunately, Microsoft has declared .NET 3.5 is supported until at least Jan 2029 (source: https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-framework). Furthermore, bringing .NET 3.5 to the game modding community can significantly increase the surface area where AsmResolver could be used.

### Additional Context

_No response_

Washi1337:
- Adds .NET 3.5 build targets for all main packages.
- Adds MonoMod.Backports as a conditional package reference for .NET 3.5 build targets. 
- Adds a few extra runtime-abstracting shims to the AsmResolver common package.
- Adds `IInputFile::BaseAddress`
- Removes `MemoryMappedDataSource`, the new implementation uses `UnmanagedDataSource`.

Closes #536


nike4613:
Backports has a few of these shims already, on `*Ex` suffixed types in the same namespace as the relevant type. (Though I see you're limiting the Backports package ref to `net35`, so they may not actually be terribly useful...)

Washi1337:
> Though I see you're limiting the Backports package ref to net35

Yes, I intend to only include MonoMod.Backports when absolutely necessary (i.e., only for the .NET 3.5 target). Using the existing shims for `Array` and `String` would therefore not work out.

src/AsmResolver/Shims/RuntimeInformationShim.cs
```diff
@@ -0,0 +1,50 @@
...
+        get;
+    }
+
+    /// <summary>
+    /// Gets a value indicating whether the current runtime runs on a Unix-based operating system.
+    /// </summary>
+    public static bool IsRunningOnUnix
+    {
+        get;
+    }
```
ds5678, line 41: I think it's more efficient to not cache the value because then the JIT can optimize branching based on runtime information.

src/AsmResolver/Shims/ArrayShim.cs
```diff
@@ -0,0 +1,27 @@
...
+#if !NET35
+    public static T[] Empty<T>() => Array.Empty<T>();
+#else
+    public static T[] Empty<T>() => ArrayHelper<T>.Empty;
+
+    private static class ArrayHelper<T>
+    {
+        public static readonly T[] Empty = new T[0];
+    }
+#endif
```
ds5678, line 24: Collection expressions might be supported on NET35. If so, I think you could use `[]` everywhere instead of having a shim.

src/AsmResolver/Shims/RuntimeInformationShim.cs
nike4613: As long as the static constructor has run when the JIT compiles a method using it, it will see this as a constant (because it gets emitted as a `static readonly` field, through a very small method body).

src/AsmResolver/Shims/ArrayShim.cs
nike4613: I think that when `Array.Empty<T>()` is not available, `[]` will always construct a new array. Probably worth double checking though.

Washi1337: This is indeed the case. On .NET 3.5 the following C#
```csharp
public static int[] ReturnEmpty() => [];
```
results in the following CIL, creating a new empty array on every call.
```
.method public static hidebysig int32[] ReturnEmpty() cil managed
{
    .maxstack 8

    IL_0000: ldc.i4.0
    IL_0001: newarr valuetype [mscorlib] System.Int32
    IL_0006: ret
}
```

src/AsmResolver/Shims/RuntimeInformationShim.cs
Washi1337: I'm pretty sure it will also depend on whether the calls to into `RuntimeInformation` are considered JIT intrinsics, which I am not sure about.

PR #544, opened 2024-03-20
MemoryMappedFile for .NET Framework 3.5

Windows10CE:


src/AsmResolver/AsmResolver.csproj
```diff
@@ -33,4 +33,8 @@
         <PackageReference Include="MonoMod.Backports" Version="1.1.0" />
     </ItemGroup>
 
+    <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
+        <PackageReference Include="System.Memory" Version="4.5.5" />
+    </ItemGroup>
```
Washi1337, line 38: I am not convinced we should do this, especially because we can easily avoid it (see other comments on unmanaged data source). This will pull in 3 additional packages for .NET standard 2.0 targets, which is not great for .NET Framework 4.x targets:

- [System.Buffers](https://www.nuget.org/packages/System.Buffers/) (>= 4.5.1)
- [System.Numerics.Vectors](https://www.nuget.org/packages/System.Numerics.Vectors/) (>= 4.4.0)
- [System.Runtime.CompilerServices.Unsafe](https://www.nuget.org/packages/System.Runtime.CompilerServices.Unsafe/) (>= 4.5.3)

src/AsmResolver/Shims/MemoryMappedFileShim.Unix.cs
```diff
@@ -0,0 +1,52 @@
...
+    private static extern bool munmap(void* adrr, nuint length);
+
+    [DllImport("libc", ExactSpelling = true, SetLastError = true)]
+    private static extern long lseek(int fd, long offset, int whence);
+
+    private void AllocateFileUnix(string path)
+    {
+        int fd = open(path, 0);
+        if (fd == -1)
+            throw new Win32Exception();
```
Washi1337, line 33: Question, do `Win32Excepion`s map onto Linux errors as well?

src/AsmResolver/Shims/MemoryMappedFileShim.Windows.cs
```diff
@@ -0,0 +1,74 @@
...
+    [DllImport("kernel32", ExactSpelling = true, SetLastError = true)]
+    [return: MarshalAs(UnmanagedType.Bool)]
+    private static extern bool CloseHandle(void* handle);
+
+    [DllImport("kernel32", ExactSpelling = true, CharSet = CharSet.Ansi, SetLastError = true)]
+    private static extern nint CreateFileMappingA(void* handle, void* mapAttributes, uint protection, uint maxSizeHigh,
+        uint maxSizeLow, sbyte* name);
+
+    [DllImport("kernel32", ExactSpelling = true, SetLastError = true)]
+    private static extern void* MapViewOfFile(void* handle, uint access, uint offsetHigh, uint offsetLow, nuint size);
```
Washi1337, line 36: The typing is inconsistent here.  The `handle` parameters are of type `void*`, while the APIs responsible for creating these handles return `nint`. Can't we just do `nint` for all handles (as is typically the convention)?

src/AsmResolver/IO/MemoryMappedInputFile.cs
```diff
@@ -21,9 +19,8 @@ public sealed class MemoryMappedInputFile : IInputFile
         public MemoryMappedInputFile(string filePath)
         {
             FilePath = filePath ?? throw new ArgumentNullException(nameof(filePath));
-            _file = MemoryMappedFile.CreateFromFile(filePath);
-            long fileSize = new FileInfo(filePath).Length;
-            _dataSource = new MemoryMappedDataSource(_file.CreateViewAccessor(0, fileSize), (ulong) fileSize);
+            _file = new MemoryMappedFileShim(filePath);
+            _dataSource = new MemoryMappedDataSource(_file);
```
Washi1337, line 23: Since we are using a raw unmanaged pointer to the data, couldn't we get rid of `MemoryMappedDataSource` alltogether and just use `UnmanagedDataSource` instead? This also removes the hard dependency for `Span<T>` in the `GetSpan` method (as we wouldn't need it anymore).

src/AsmResolver/Shims/MemoryMappedFileShim.Windows.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace AsmResolver.Shims;
+
+internal sealed unsafe partial class MemoryMappedFileShim
+{
+    private const uint GENERIC_READ = 0x80000000;
+    private const uint FILE_SHARE_READ = 0x2;
+    private const uint OPEN_EXISTING = 3;
+    private const uint PAGE_READONLY = 2;
+    // not on msdn, stole from terrafx
+    private const uint FILE_MAP_READ = 4;
```
Washi1337, line 15: "stole"? Maybe add a reference?

Washi1337 (changes requested):

src/AsmResolver/Shims/MemoryMappedFileShim.Unix.cs
Windows10CE: Yes in CoreCLR (and newer Mono by proxy). Legacy Mono (aka Unity) does whatever https://github.com/Unity-Technologies/mono/blob/e5c9e6e3feab0841f26dd6df5e19c0fe5ab0105d/mcs/class/System/ReferenceSources/Win32Exception.cs#L9 is, which I don't know is right.

src/AsmResolver/Shims/MemoryMappedFileShim.Windows.cs
Windows10CE: My logic was pretty much
 - if the invalid/error return is -1, use nint so it can be compared to that
- else if the invalid return is null, just use void*
if you'd like to me to standardize on one or the other, I don't mind either

Washi1337 (approved):
Awesome, thanks a lot for looking into it!

PR #555, opened 2024-05-12
Set Accessors for Semantic Methods

ds5678:
Should these set accessors modify `MethodDefinition.Semantics` too?

### DIFF
--- a/src/AsmResolver.DotNet/EventDefinition.cs
+++ b/src/AsmResolver.DotNet/EventDefinition.cs
@@ -139,20 +139,29 @@ namespace AsmResolver.DotNet
         /// <summary>
         /// Gets the method definition representing the first add accessor of this event definition.
         /// </summary>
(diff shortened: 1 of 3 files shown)


Washi1337:
This _should_ be already handled by [MethodSemanticsCollection](https://github.com/Washi1337/AsmResolver/blob/master/src/AsmResolver.DotNet/Collections/MethodSemanticsCollection.cs). If not, then this should be considered a bug.

ds5678:
> This _should_ be already handled by [MethodSemanticsCollection](https://github.com/Washi1337/AsmResolver/blob/master/src/AsmResolver.DotNet/Collections/MethodSemanticsCollection.cs). If not, then this should be considered a bug.

This is correct. I added a unit test to make sure.

Washi1337:
Thanks for verifying!

PR #557, opened 2024-05-19
Address several issues with reading specially crafted .NET metadata

ElektroKill:
This PR fixes a couple of issues with reading specially crafted files that abuse implementation quirks of the CLR metadata reader.

The following quirks were addressed (sample files can be found in the PR):
- The presence of the `#Schema` stream forces the parsing of metadata as if it was EnC metadata even if the tablees stream name suggests compressed metadata.
- When reading EnC metadata, the CLR performs a case insensitive comparison when looking for stream names`#Blob`, `#US`, `#Strings`, `#GUID` as compared to byte for byte compare when reading compressed metadata.
- When reading EnC metadata, the CLR checks for the presence of a stream named `#JTD` and if it found it forces all column index sizes to 4 bytes.

All the above-mentioned quirks were addressed in the PR and have accompanying tests.


src/AsmResolver.PE/DotNet/Metadata/SerializedMetadata.cs
```diff
@@ -89,8 +104,15 @@ protected override IList<IMetadataStream> GetStreams()
             if (_streamHeaders.Length == 0)
                 return base.GetStreams();
 
+            var flags = MetadataStreamReaderFlags.None;
+            if (IsEncMetadata)
+                flags |= MetadataStreamReaderFlags.IsEnc;
```
Washi1337, line 109: This has a data race, as `IsEncMetadata` could be set before the streams are initialized. We probably want to add a `_isEncMetadata` field that stays private to the `SerializedMetadata` class,  similar to `_hasJtdStream`.

test/AsmResolver.PE.Tests/DotNet/Metadata/MetadataTest.cs
```diff
@@ -198,5 +203,51 @@ public void SelectFirstUserStringsStreamInEnCMetadata()
...
+            Assert.True(tablesStream.ForceLargeColumns);
+
+            for (var i = TableIndex.Module; i <= TableIndex.GenericParamConstraint; i++)
+                Assert.Equal(IndexSize.Long, tablesStream.GetTableIndexSize(i));
+
+            for (var i = TableIndex.Document; i <= TableIndex.CustomDebugInformation; i++)
+                Assert.Equal(IndexSize.Long, tablesStream.GetTableIndexSize(i));
+
+            for (var i = CodedIndex.TypeDefOrRef; i <= CodedIndex.HasCustomDebugInformation; i++)
+                Assert.Equal(IndexSize.Long, tablesStream.GetIndexEncoder(i).IndexSize);
```
Washi1337, line 246: Use `Assert.All` instead of loops.

test/AsmResolver.PE.Tests/DotNet/Metadata/MetadataTest.cs
```diff
@@ -198,5 +203,51 @@ public void SelectFirstUserStringsStreamInEnCMetadata()
...
+            var metadata = peImage.DotNetDirectory!.Metadata!;
+
+            Assert.True(metadata.TryGetStream(out BlobStream? _));
+            Assert.True(metadata.TryGetStream(out GuidStream? _));
+            Assert.True(metadata.TryGetStream(out StringsStream? _));
+            Assert.True(metadata.TryGetStream(out UserStringsStream? _));
+        }
+
+        [Fact]
+        public void UseLargeTableIndicesWhenJTDStreamIsPresentInEnCMetadata()
```
Washi1337, line 230: This also needs a writer test that validates the right column sizes are used (e.g., reread rebuilt binary and test column sizes).

test/AsmResolver.PE.Tests/DotNet/Metadata/MetadataTest.cs
```diff
@@ -198,5 +203,51 @@ public void SelectFirstUserStringsStreamInEnCMetadata()
...
+        {
+            var peImage = PEImage.FromBytes(Properties.Resources.HelloWorld_SchemaStream);
+            AssertCorrectStreamIsSelected<BlobStream>(peImage, true);
+            AssertCorrectStreamIsSelected<GuidStream>(peImage, true);
+            AssertCorrectStreamIsSelected<StringsStream>(peImage, true);
+            AssertCorrectStreamIsSelected<UserStringsStream>(peImage, true);
+        }
+
+        [Fact]
+        public void UseCaseInsensitiveComparisonForHeapNamesInEnCMetadata()
```
Washi1337, line 218: We probably want a counterpart test that tests whether case sensitivity is used for non-enc metadata (e.g., two binaries with both a `#Strings` and a `#strings` stream, one EnC and one non-EnC metadata).

Washi1337 (changes requested):

PR #575, opened 2024-07-12
Add more nullable attributes to MethodDefinition and TypeDefinition

ds5678:


Washi1337:
I am unfamiliar with `PolySharp`. What was the rationale for switching to it?

nike4613:
> I am unfamiliar with `PolySharp`. What was the rationale for switching to it, and, more importantly, does it behave similar to Nullable in the sense that it won't introduce a new dependency?

It's a more complete polyfill library, exposing basically all attributes that are normally defined in the BCL when they're not available in a target. They're injected at the source level (via a Roslyn SG I believe, to avoid emitting attributes available in referenced assemblies), so no new assembly dependencies.

Washi1337:
Inspecting the AppVeyor builds in a decompiler, it seems that Polysharp added quite a few extra attributes that are actually not used within AsmResolver (e.g., `[StringSyntax]`, `[AsyncMethodBuilder]`...), especially for target frameworks like netstandard2.0. It seems that Polysharp simply looks at the target runtime and includes the source if it doesn't exist, rather than basing this on actual usage of the attribute. Is there an option to configure this?

ds5678:
> It seems that Polysharp simply looks at the target runtime and includes the source if it doesn't exist, rather than basing this on actual usage of the attribute.

That's how Nullable works too. Detecting usage is very difficult.

> Is there an option to configure this?

These are the configuration options. https://github.com/Sergio0694/PolySharp?tab=readme-ov-file#options-%EF%B8%8F

Washi1337:
> That's how Nullable works too. Detecting usage is very difficult.

The difference is that Nullable only includes the nullable attributes, which are used by AsmResolver. PolySharp, on the other hand, includes many more attributes not used by AsmResolver. I want to avoid the unnecessary inclusion of unused attributes.

ds5678:
Is the binary size difference noticeable?

Washi1337:
> Is the binary size difference noticeable?

Even if the final output size is unnoticeable, I still do not really see a reason to switch to a source-gen library that effectively does the same and includes the same nullable attributes as the one we are using right now, but adds more attributes/clutter we don't use. Unless there is some benefit to using PolySharp compared to Nullable, isn't that a net negative regardless?

ds5678:
I think Nullable is lacking the "Member" nullable attributes.

Washi1337:
We are already using `MemberNotNullWhen` via the Nullable package for various properties (see 
https://github.com/search?q=repo%3AWashi1337%2FAsmResolver+membernotnullwhen&type=code).

ds5678:
I thought it wouldn't matter, but it's not that important to me, so I'll just change it back when I have time.

Washi1337:
Thanks!

PR #581, opened 2024-09-09
SentinelTypeSignature as a singleton

ds5678:


Washi1337:
Note: Ignoring failing tests because AppVeyor recently dropped .NET 5.0 support in their builder image.

PR #582, opened 2024-09-11
GitHub Actions Workflow

Aragas:
The flow is like this:
1. Build Release x86/x64 binaries on Windows. Artifacts folder published as artifacts
2. When testing on Windows/Ubuntu/OSX we download the Windows binaries and run the tests
3. Test result are aggregated and published | NuGet Packages are published
4. Cleanup of intermediate files

We have two types of tests:
* Non Experimental. Windows x86/x64 - mandatory for tests to pass
* Experimental. Ubuntu x64, OSX 13 x64, OSX 14 arm64/x64 Rosetta2 - will have their own test result so we can differentiate expected failures from non expected ones. Any test failures will not affect the workflow result 
  
Completed:
* Windows x86 support
* NuGet publish on master push
* Nighly NuGet publish on development push
* PR's to master/development only publish the artifacts folder
* OS X arm64 support
* OS X x64 on arm64 via Rosetta2
* Wine on Ubuntu
  
Not completed:
* Wine on OSX (looks to be setup somewhat correct, but tests fail where they should not)


.github/workflows/test-and-publish.yml
```diff
@@ -0,0 +1,120 @@
...
+          echo "FRAMEWORK_TARGET_FRAMEWORKS=${{env.FRAMEWORK_TARGET_FRAMEWORKS}}" >> $GITHUB_OUTPUT
+
+  build:
+    name: Perform Build
+    needs: [variables]
+    strategy:
+      matrix:
+        runner: [ 
+          { image: 'windows-latest', arch: 'x64' },
+          { image: 'windows-latest', arch: 'x86' },
```
Washi1337, line 79: Do we need two separate builds?

.github/workflows/test-and-publish.yml
```diff
@@ -0,0 +1,120 @@
...
+  upload-event-file:
+    name: Upload Event File
+    runs-on: ubuntu-latest
+    steps:
+      - name: Upload event file
+        uses: actions/upload-artifact@v4
+        with:
+          name: test-event-file
+          path: ${{ github.event_path }}
+          retention-days: 1
```
Washi1337, line 51: Unsure about this, should these be exposed to the public?

.github/workflows/test-and-publish.yml
```diff
@@ -0,0 +1,120 @@
...
+        ]
+        build_configuration: ${{fromJson(needs.variables.outputs.BUILD_CONFIGURATIONS)}}
+    uses: ./.github/workflows/reusable-test.yml
+    with:
+      image: ${{matrix.runner.image}}
+      architecture: ${{matrix.runner.arch}}
+      dotnet_target_frameworks: ${{needs.variables.outputs.DOTNET_TARGET_FRAMEWORKS}}
+      build_configuration: ${{matrix.build_configuration}}
+      #ignore non-windows tests since support is yet lacking
+      is_experimental: ${{matrix.runner.image != 'windows-latest'}}
```
Washi1337, line 110: 713853a30838d753f02fe8284ef27e5201a68015 should make it possible to change this to macos only.

.github/workflows/reusable-test.yml
```diff
@@ -0,0 +1,157 @@
...
+        run: '& "${{steps.get-mono-path.outputs.path}}" --version'
+        shell: pwsh
+      # Mono
+
+      - name: Download Build Cache
+        uses: ./.github/actions/test-build-cache-download
+        with:
+            #image: ${{inputs.image}}
+            image: 'windows-latest'
+            architecture: ${{(inputs.architecture == 'arm64' && 'x64') || inputs.architecture}}
```
Washi1337, line 132: I don't get this, what does this do? Shouldn't we be able to just pull the build cache from the same build?

.github/workflows/reusable-build.yml
```diff
@@ -0,0 +1,97 @@
...
+          script: |
+            let gitHash = "${{ steps.git-vars.outputs.git_hash }}";
+            let runNumber = "${{ github.run_number }}"; 
+            let rawGitRef = "${{ steps.git-vars.outputs.git_branch }}";
+            console.log("rawGitRef: " + rawGitRef);
+            let gitRef = rawGitRef.replace(/^refs\/heads\//, "");
+            if(gitRef.indexOf("refs/pull/") === 0) {
+              gitRef = "pr-" + gitRef.substring(10, gitRef.lastIndexOf("/"));
+            }
+            var versSuffix = `${gitRef}.${runNumber}+${gitHash}`;
```
Washi1337, line 55: Will this go fine if the branch has the form `feature/xyz`?

Washi1337 (commented):
Few things I am unsure about/would like to have some clarification on, but other than that LGTM.

Also, if we're letting github actions deploy to nuget, we should remove the nuget deploy tag from `appveyor.yml`

.github/workflows/test-and-publish.yml
nike4613: The event file doesn't have any private information, it's just a webhook payload.

.github/workflows/test-and-publish.yml
Aragas: For building the C++ projects

.github/workflows/reusable-test.yml
Aragas: We are, but the only have Windows x86/x64 build caches. If we have OSX Arm64, it falls back to Windows x64

.github/workflows/test-and-publish.yml
Aragas: Done, Ubuntu not passes tests

.github/workflows/reusable-build.yml
Aragas: Looks like there are plentiful of special characters we need to replace. I've added support of replacing `/`, `!`, `@`, `#`, `$`, `%`, `&` with `_`

PR #590, opened 2024-10-19
Windows x86 build fix

Aragas:


ds5678:
Was targeting master intentional?

Washi1337:
This seems to be no longer required.

PR #617, opened 2025-03-02
Use public API to get type from address where available.

teo-tsirpanis:
.NET 7 has introduced the [`RuntimeTypeHandle.FromIntPtr`](https://learn.microsoft.com/en-us/dotnet/api/system.runtimetypehandle.fromintptr) API, and this PR uses it in `DynamicTypeSignatureReader` where available, instead of doing private reflection.


Washi1337:
Interesting, this is a nice addition of .NET 7. Thanks for incorporating it.

Could you please rebase and change the merge target of the PR to `development`. Thanks!

teo-tsirpanis:
Done.

Washi1337:
Thanks!

PR #619, opened 2025-03-15
Use inbox memory-mapped file APIs where available.

teo-tsirpanis:
This PR makes the memory-mapped file shim implementation added in #544 to be used only when targeting .NET Framework 3.5. For all newer frameworks, the `MemoryMappedFileShim` class' implementation simply forwards to the inbox framework APIs.

We also multi-target to .NET Framework 4.6.2 (the earliest supported 4.x version) in order to avoid the shims where we can.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -3,7 +3,7 @@
     <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
 
     <PropertyGroup>
-        <TargetFrameworks>net8.0;net6.0;net35;netcoreapp3.1;netstandard2.0;netstandard2.1</TargetFrameworks>
+        <TargetFrameworks>net8.0;net6.0;net35;net462;netcoreapp3.1;netstandard2.0;netstandard2.1</TargetFrameworks>
         <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
         <Nullable>enable</Nullable>
         <IsTrimmable>true</IsTrimmable>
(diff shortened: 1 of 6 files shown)


teo-tsirpanis:
> Is there a way to avoid the inclusion of `System.Runtime.InteropServices.RuntimeInformation` for this new target?

We can target `net472` as well where `RuntimeInformation` is inbox. Or we can target only `net472`, but earlier frameworks will use the assemblies compiled for `net35`. I think it's better to target both, to provide a consistent experience with .NET Standard.

Washi1337:
> but earlier frameworks will use the assemblies compiled for net35

That is wild, I was not aware of this. I expected a fw 4.6.2 project to compile against the netstandard 2.0 package, as it is the highest compatible version. Thanks for pointing it out. Targeting both then indeed makes sense in my opinion.

Washi1337:
@teo-tsirpanis could you add the `net472` build target as well? Then I will merge the PR :)

src/Directory.Build.props
```diff
@@ -3,7 +3,7 @@
     <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
 
     <PropertyGroup>
-        <TargetFrameworks>net8.0;net6.0;net35;netcoreapp3.1;netstandard2.0;netstandard2.1</TargetFrameworks>
+        <TargetFrameworks>net8.0;net6.0;net35;net462;netcoreapp3.1;netstandard2.0;netstandard2.1</TargetFrameworks>
```
teo-tsirpanis, line 6:
```suggestion
        <TargetFrameworks>net8.0;net6.0;net462;net472;net35;netcoreapp3.1;netstandard2.0;netstandard2.1</TargetFrameworks>
```

teo-tsirpanis:
@Washi1337 done.

Washi1337:
Thanks!

PR #621, opened 2025-03-23
Fix CorLib module self-reference

### LINKED ISSUE #620: Invalid `CorLibTypeFactory::CorLibScope` when creating corlib module
### AsmResolver Version

6 beta 3

### .NET Version

.NET 9

### Operating System

Windows

### Describe the Bug

When attempting to create a corlib module, `CorLibTypeFactory::CorLibScope` should be the `ModuleDefinition`, but it's actually an assembly reference to itself.

This self-reference prevents Roslyn from recognizing the assembly as a valid corlib to compile against.

### How To Reproduce

```cs
AssemblyDefinition assembly = new("mscorlib", new Version(4, 0, 0, 0));
ModuleDefinition module = new("mscorlib", new AssemblyReference(assembly));
assembly.Modules.Add(module);
Assert.That(module.CorLibTypeFactory.CorLibScope is ModuleDefinition);
```

### Expected Behavior

`module.CorLibTypeFactory.CorLibScope is ModuleDefinition`

### Actual Behavior

`module.CorLibTypeFactory.CorLibScope is AssemblyReference`

### Additional Context

My current workaround:

```cs
private sealed class CorLibModuleDefinition : ModuleDefinition
{
    public CorLibModuleDefinition(string? name, AssemblyDefinition assembly) : base(name, new(assembly))
    {
        CorLibTypeFactory = new(this);
        AssemblyReferences.Clear();
    }
}
```

~~My proposed solution is to change the `ModuleDefinition` constructor from `(string?, AssemblyReference)` to `(string?, IResolutionScope)`~~ This wouldn't work.

ds5678:
Resolves #620

### DIFF
--- a/src/AsmResolver.DotNet/ModuleDefinition.cs
+++ b/src/AsmResolver.DotNet/ModuleDefinition.cs
@@ -314,19 +314,16 @@ namespace AsmResolver.DotNet
         /// Defines a new .NET module.
         /// </summary>
         /// <param name="name">The name of the module.</param>
-        /// <param name="corLib">The reference to the common object runtime (COR) library that this module will use.</param>
-        public ModuleDefinition(string? name, AssemblyReference corLib)
+        /// <param name="corLib">
+        /// The reference to the common object runtime (COR) library that this module will use.
+        /// If null, this module will be treated as a standalone core library.
+        /// </param>
+        public ModuleDefinition(string? name, AssemblyReference? corLib)
             : this(new MetadataToken(TableIndex.Module, 0))
         {
             Name = name;
 
-            // CorLib libraries should not reference themselves.
-            // https://github.com/Washi1337/AsmResolver/issues/620
-            IResolutionScope corLibScope = corLib.Name == name && name is not null && KnownCorLibs.KnownCorLibNames.Contains(name)
-                ? this
-                : corLib.ImportWith(DefaultImporter);
-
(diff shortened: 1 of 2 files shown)


Washi1337:
I am not sure about the proposed solution.

A normal C# standard compiler sets the module name to the actual file names, including their file extension. For example, an assembly `MyAssembly` will have a module called `MyAssembly.exe` or `MyAssembly.dll`. Thus, `mscorlib`'s module is actually called `mscorlib.dll`. To make matters worse, in legacy .NET Framework it is actually called `CommonLanguageRuntimeLibrary`, a name that doesn't match any corlib assembly name nor does it have an extension (hip hip hurray for inconsistent legacy :upside_down_face:).

Should we solely rely on the input name or scope to determine all this? I can foresee the problem of ambiguity which has direct implications on which underlying corlib scope and metadata resolver that is created. Does the name `mscorlib` imply we want to create a mscorlib for .NET Framework (and thus doesn't require an assembly resolver) or do we want to make a forwarder assembly as is happening in .NET Core/.NET (which _does_ require a specific resolver)?

Would it make more sense to have it explicitly specified by the caller (e.g., maybe a constructor overload or factory method)?

ds5678:
Explicit specification by the caller does sound like the better option, given the issues you've outlined. How do you think the caller should specify it? I feel like a static factory method might be best.

PR #624, opened 2025-04-20
Fix fields and methods being accessible from nested types

ds5678:


Washi1337:
Thanks for the catch!

PR #628, opened 2025-05-09
Respect RVA field alignment requirements

Windows10CE:
https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md#rules-for-il-rewriters

### DIFF
--- a/docs/guides/dotnet/importing.md
+++ b/docs/guides/dotnet/importing.md
@@ -86,15 +86,15 @@ method instead.
 
 Types and members can also be imported by passing on an instance of
 various `System.Reflection` classes.
-             
-|Member type to import |Method to use        |Result type          |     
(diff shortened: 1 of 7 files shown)


Washi1337:
Small note, but we should probably extract this alignment logic so it can also be used in `UnmanagedPEFileBuilder`. Other than that, LGTM

https://github.com/Washi1337/AsmResolver/blob/07d02442bbde97a77f98d538814ff9382bec7a53/src/AsmResolver.PE/Builder/UnmanagedPEFileBuilder.cs#L650-L682

Windows10CE:
Adding this as reference for the same thing being done in Cecil, for anyone looking back at this PR: https://github.com/jbevain/cecil/pull/817

Washi1337:
Thanks!

PR #632, opened 2025-05-19
Use case insensitive compare when looking for JTD stream in ENC metadata

ElektroKill:
Addresses an oversight I made in https://github.com/Washi1337/AsmResolver/pull/557.
It appears that CLR also uses case-insensitive comparison for the `#JTD` stream in ENC metadata.


Washi1337:
Could we add a unit test for this behavior, to avoid future regressions?

Washi1337:
Thanks!

PR #646, opened 2025-06-11
Include assembly spec for 'corelib' in attributes

Sergio0694:
This fixes the custom attribute serialization logic to emit FQNs with assembly specs in all cases.

Context:

image

image

### DIFF
--- a/src/AsmResolver.DotNet/Signatures/Parsing/CustomAttributeArgumentWriter.cs
+++ b/src/AsmResolver.DotNet/Signatures/Parsing/CustomAttributeArgumentWriter.cs
@@ -91,7 +91,15 @@ namespace AsmResolver.DotNet.Signatures.Parsing
 
             if (argumentType.IsTypeOf("System", "Type"))
             {
-                writer.WriteSerString(TypeNameBuilder.GetAssemblyQualifiedName((TypeSignature) element));
+                // Type names must include the assembly spec for 'corelib' as well. E.g., this is what Roslyn emits:
+                //
+                // 'System.Collections.Generic.List`1[[System.Uri, System.Runtime, Version=10.0.0.0, Culture=neutral,
+                //  PublicKeyToken=b03f5f7f11d50a3a]], System.Collections, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
+                //
+                // When parsing 'Type' attribute elements, if the assembly spec is not specified, types would otherwise default to 'corlib', but that
+                // would not necessarily be correct. For instance, 'System.Uri' is not in 'System.Private.CoreLib.dll', it's in 'System.Private.Uri.dll'.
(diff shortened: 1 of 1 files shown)


Windows10CE:
I believe that the default for this method should be changed to not omit corlib, as there are a few other usages in the codebase (security attributes, custom MarshalAs marshallers, safe arrays) that I believe are also broken in this same way.

By that I mean `_omitCorLib` should be removed from this class entirely, and all relevant overloads removed. This would match us to the behavior that Roslyn has today.

nike4613:
To cite ECMA-335:

§ II.23.3
> - If the parameter kind is System.Type, (also, the middle line in above diagram) its
> value is stored as a SerString (as defined in the previous paragraph), representing its
> canonical name. The canonical name is its full type name, followed optionally by
> the assembly where it is defined, its version, culture and public-key-token. If the
> assembly name is omitted, the CLI looks first in the current assembly, and then in
> **the system library (mscorlib)**; in these two special cases, it is permitted to omit the
> assembly-name, version, culture and public-key-token.

(emphasis mine)

There is no amendment for .NET Core that updates this text, but given .NET Core's reference assembly split, it is unclear what exactly it means. *In practice*, as @Windows10CE says in the screenshots, the runtime looks in `System.Private.CoreLib`, and NOT the .NET Core `System.Runtime` reference assembly.

Washi1337:
Unsure what this PR really fixes to be honest, given the lack of context or new unit tests that describe the bug/issue.

If I understand the problem correctly, it seems AsmResolver incorrectly decides to omit the corlib assembly name in a type, because the type does not reside in a "proper" corlib (i.e., defined in `System.Private.Corlib`) but in a corlib "facade" (`System.Runtime`)? Do we have any evidence backing this up in either the source of `dotnet/runtime` or else binaries that indicate this behavior?

If that is indeed true, then this PR feels like a band-aid solution, or at least an incomplete one, to the underlying core problem (the distinction between "proper" and "facade" corlibs). Just because we are choosing to always write external scopes, does not mean that this is what other processors may do.

e.g., if we do actually encounter a(n incorrect) `System.Uri` with no assembly spec in a .NET Core binary, the current approach then tries to look for it in the wrong corlib assembly as well, because a normal .NET Core binary references `System.Runtime` and not `System.Private.Corlib`. Thus, `ModuleDefinition.CorLibTypeFactory.CorLibScope` will not contain the right resolution context and happily forward to `System.Private.Uri`. Resolution will succeed when in reality it should fail.

https://github.com/Washi1337/AsmResolver/blob/db9b208d1591f51cb1040e775010f3d9f7bfa737/src/AsmResolver.DotNet/Signatures/Parsing/TypeName.cs#L19-L36

I am not opposed to always emitting the assembly spec if scope != current module, but we should still ensure that this behavior is consistent with all runtimes, for both writing and reading assemblies. We could also consider actually introducing the distinction between proper and facade corlibs, and adjust `TypeNameBuilder` accordingly.

Sergio0694:
Closing this in favor of #648.

PR #648, opened 2025-06-12
Fix not fully qualified type names in blobs

Windows10CE:
This PR supersedes #646.

Created as a draft as I still need to figure out some tests and need the following question answered: upon failing to resolve from what we assume to be the implementing core library for the target runtime (or if we cannot pick one, in the case of netstandard), what do we do? The implementation I have started this PR with does a second layer of fallback to CorLibScope and attempts a third resolve there, but this means AsmResolver could succeed resolution in more cases than it should. We could do nothing, but this means non-FQ type names to corelib types in netstandard would never be able to resolve without user intervention. If there are any other ideas for how to handle these types, it would be great to hear them, but I think no matter what AsmResolver is going to end up with some odd behavior here (since I don't believe definitive correct behavior exists).

### DIFF
--- a/src/AsmResolver.DotNet/KnownCorLibs.cs
+++ b/src/AsmResolver.DotNet/KnownCorLibs.cs
@@ -262,7 +262,7 @@ namespace AsmResolver.DotNet
 
         static KnownCorLibs()
         {
-            KnownCorLibReferences = new HashSet<AssemblyReference>(new SignatureComparer())
+            KnownCorLibReferences = new HashSet<AssemblyReference>(SignatureComparer.Default)
             {
                 NetStandard_v2_0_0_0,
                 NetStandard_v2_1_0_0,
@@ -313,6 +313,26 @@ namespace AsmResolver.DotNet
             throw new ArgumentException($"Invalid or unsupported runtime version {runtimeInfo}.");
         }
 
+        /// <summary>
+        /// Obtains a reference to the default core lib reference for the provided .NET target runtime.
+        /// </summary>
+        /// <param name="runtimeInfo">The runtime to target.</param>
+        /// <returns>The reference to the default core lib.</returns>
+        /// <exception cref="ArgumentException">The runtime information is invalid or unsupported.</exception>
+        public static AssemblyReference? TryImplFromRuntimeInfo(DotNetRuntimeInfo runtimeInfo)
+        {
+            if (runtimeInfo.IsNetStandard)
+                return null;
+
+            if (runtimeInfo.IsNetFramework)
+                return SelectFrameworkCorLib(runtimeInfo.Version);
+
+            if (runtimeInfo.IsNetCoreApp)
+                return SelectNetCoreImplCorLib(runtimeInfo.Version);
+
+            throw new ArgumentException($"Invalid or unsupported runtime version {runtimeInfo}.");
+        }
+
         private static AssemblyReference SelectFrameworkCorLib(Version version) => version.Major < 4
             ? MsCorLib_v2_0_0_0
             : MsCorLib_v4_0_0_0;
@@ -345,6 +365,20 @@ namespace AsmResolver.DotNet
                 (8, 0) => SystemRuntime_v8_0_0_0,
                 (9, 0) => SystemRuntime_v9_0_0_0,
                 (10, 0) => SystemRuntime_v10_0_0_0,
+                _ => throw new ArgumentException($"Invalid or unsupported .NET or .NET Core version {version}.")
+            };
+        }
+
+        private static AssemblyReference SelectNetCoreImplCorLib(Version version)
+        {
+            return (version.Major, version.Minor) switch
+            {
+                (1, 0 or 1) or (2, 0 or 1) or (3,0 or 1) => SystemPrivateCoreLib_v4_0_0_0,
+                (5, 0) => SystemPrivateCoreLib_v5_0_0_0,
+                (6, 0) => SystemPrivateCoreLib_v6_0_0_0,
+                (7, 0) => SystemPrivateCoreLib_v7_0_0_0,
+                (8, 0) => SystemPrivateCoreLib_v8_0_0_0,
+                (9, 0) => SystemPrivateCoreLib_v9_0_0_0,
                 _ => throw new ArgumentException($"Invalid or unsupported .NET or .NET Core version {version}.")
             };
         }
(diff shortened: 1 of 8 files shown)


Windows10CE:
For some additional references for behavior being changed by this PR, all checked 2025-06-12:

ECMA-335 6th Edition, Section II.23.3, bottom of page 268, describes how to do resolution when no assembly spec is given for Type arguments:
https://ecma-international.org/wp-content/uploads/ECMA-335_6th_edition_june_2012.pdf

The current behavior of CoreCLR, which attempts current assembly, then `typeof(object).Assembly` (System.Private.CoreLib), then the TypeResolve event of the current AssemblyLoadContext:
https://github.com/dotnet/runtime/blob/b49ebac33e06556804af96afebbb8f706d51a2e2/src/coreclr/System.Private.CoreLib/src/System/Reflection/TypeNameResolver.CoreCLR.cs#L344

The current behavior of Roslyn (the C#/VB compiler), which always writes FQNs when there is an assembly being referenced: https://github.com/dotnet/roslyn/blob/2f32bf87e211fe5319c94f442c4ad56254bc75e0/src/Compilers/Core/Portable/PEWriter/TypeNameSerializer.cs#L194

Cecil reading always uses `module.TypeSystem.CoreLibrary` when it cannot find a type definition (the current behavior of AsmResolver):
https://github.com/jbevain/cecil/blob/3136847ea620fb9b4a3ff96bc4f573148e8bd2e4/Mono.Cecil/TypeParser.cs#L356

Cecil writing follows the ECMA section to the letter and only omits it when the assembly being referenced is exactly current module or mscorlib (this behavior is absolutely incorrect and almost certainly has real issues, I just cannot be bothered to synthesize examples of it, this fails when referencing mscorlib for any type that is not today in System.Private.CoreLib and then loading that attribute with CoreCLR):
https://github.com/jbevain/cecil/blob/3136847ea620fb9b4a3ff96bc4f573148e8bd2e4/Mono.Cecil/TypeParser.cs#L521

dnLib reading does the same behavior as current AsmResolver and Cecil, attempting current module, then the referenced corlib:
https://github.com/0xd4d/dnlib/blob/c78d296c522aae0520df2afd825d48266321cf36/src/DotNet/CustomAttributeReader.cs#L24

dnLib writing has an option flag for whether to enable the same optimization that AsmResolver currently employs:
https://github.com/0xd4d/dnlib/blob/c78d296c522aae0520df2afd825d48266321cf36/src/DotNet/Writer/Metadata.cs#L182-L196

src/AsmResolver.DotNet/KnownCorLibs.cs
```diff
@@ -326,5 +346,19 @@ private static AssemblyReference SelectNetCoreCorLib(Version version)
...
+        private static AssemblyReference SelectNetCoreImplCorLib(Version version)
+        {
+            return (version.Major, version.Minor) switch
+            {
+                (1, 0 or 1) or (2, 0 or 1) or (3,0 or 1) => SystemPrivateCoreLib_v4_0_0_0,
+                (5, 0) => SystemPrivateCoreLib_v5_0_0_0,
+                (6, 0) => SystemPrivateCoreLib_v6_0_0_0,
+                (7, 0) => SystemPrivateCoreLib_v7_0_0_0,
+                (8, 0) => SystemPrivateCoreLib_v8_0_0_0,
+                (9, 0) => SystemPrivateCoreLib_v9_0_0_0,
```
Sergio0694, line 358: Add .NET 10 too while at it? Added it in #645.

Windows10CE:
I believe for this to work as well as possible, RuntimeContext needs some way to query for what the corelib assembly is for that context, not tied to anything that the assembly itself is targeting. That way scenarios such as an assembly targeting netstandard loaded into RuntimeContext with net8.0 assemblies will work as expected, instead of only being able to do a non-configurable guess at what the correct assembly is based on hardcoded known core libraries. I am not really confident on how adding that to RuntimeContext should work, though.

nike4613:
Further discussion suggests that basically all of the rules @Windows10CE listed are wrong, and the only "correct" one to use for writing is this:

> The only types (aside from those in the current assembly) which may have their assembly name omitted are those listed in the ECMA-335 `CILLibraryTypes.xml` supplement as being BOTH in the assembly `mscorlib` AND a part of either the library `BCL` or `RuntimeInfrastructure` (i.e. the Kernel profile).

The library requirement could likely be relaxed, though I'm not sure how safe that is. If we *really* cared, we could precompute the full set of types that qualify, but its better to just always specify when writing. Beyond that, .NET Core considers `System.Private.CoreLib` to be its "system library", so it searching there is correct (and AsmResolver should too, when resolving).

Windows10CE:
@Washi1337 Do you have anything you'd add to or change about this PR (other than unit tests)? I'm not completely sure how I'm going to go about testing the reading functionality at the moment, so if you have any particular advice about that I would also appreciate it

Washi1337:
> @Washi1337 Do you have anything you'd add to or change about this PR (other than unit tests)? 

LGTM

> I'm not completely sure how I'm going to go about testing the reading functionality at the moment, so if you have any particular advice about that I would also appreciate it

`TypeNameParser` can be tested on its own, as long as you give it a context module (with the appropriate runtime context). For this change I think it is sufficient to check that if the assembly spec is omitted, it will behave as the runtime does. That is, it will prefer the module itself first (e.g., `DummyModule.Program` is resolvable), then the assumed runtime's impl corlib (e.g., `System.Int32` is resolvable), and also not resolve using the reference corlib (e.g., `System.Uri` should not be resolvable) even if `System.Runtime` is used as corlib scope.

nike4613:
It might be a good idea to include, somewhere in the XMLdocs, a brief explanation on why there is (and needs to be) a distinction between a module's corlib and a runtime "assumed" corlib. It's not at all obvious, even to someone reasonably familiar with the space, unless they've previously come across this exact issue.

Windows10CE:
It might deserve its own subsection in https://github.com/Washi1337/AsmResolver/blob/development/docs/guides/dotnet/advanced-module-reading.md ? Either that or a subsection in a larger article about what RuntimeContext is, why it exists, and how to use it

Washi1337 (approved):

Washi1337:
@Windows10CE Currently, RuntimeContext is briefly described as a subsection in https://github.com/Washi1337/AsmResolver/blob/master/docs/guides/dotnet/basics.md#opening-multiple-net-modules-using-runtime-contexts, but we could indeed make it a dedicated page. Advanced module reading probably is best reserved for actually customizing the parsers etc. Do you plan on adding docs changes as well? I am also fine with delaying this to a later time / putting it in a separate PR.

Windows10CE:
I'm happy to put doc changes in a separate PR to get this one merged, I don't want to hold up a beta release on doc changes 😅

Sergio0694:
+1 to this. If possible, it'd be nice to merge this first, so that as soon as this one is in, we can also resume testing things in cswinrtgen and help validate that these changes also do address all the various scenarios we had issues with 😅

PR #650, opened 2025-06-21
Fix Utf8String concatenation for non-ascii strings

ds5678:


src/AsmResolver/Utf8String.cs
```diff
@@ -200,7 +200,7 @@ public Utf8String Concat(byte[]? other)
             if (other is null || other.Length == 0)
                 return this;
 
-            byte[] result = new byte[Length + other.Length];
+            byte[] result = new byte[_data.Length + other.Length];
```
Washi1337, line 203: Looking at this now, shouldn't these actually be both `ByteCount` instead of `Length`?

ds5678: `other` is a byte array, so no.

Washi1337: Ah yes, my bad. For some reason I was under the impression we were editing `Concat(Utf8String)`.

PR #656, opened 2025-07-19
Fix double ownership of nested serialized type definitions

Windows10CE:
Currently, code that does something like:
```cs
var module = ModuleDefinition.FromFile(@"C:\Users\user\RiderProjects\ConsoleApp1\Test\bin\Debug\net9.0\Test.dll");
var type = module.TopLevelTypes.Single(it => it.Name == "Class1");
var nestedType = type.NestedTypes[0];
type.NestedTypes.Remove(nestedType);
module.TopLevelTypes.Add(nestedType);
```
would throw an exception because the constructor for SerializedTypeDefintion would set _module (through IOwnedCollectionElement) even when the type definition is not "owned" by the module, but instead by its declaring type. Changing it to only do so when no parent type exists fixes the issue.


Washi1337:
Thanks!

PR #664, opened 2025-08-18
Support reading from streams

ds5678:


Washi1337:
`StreamDataSource` needs to use locking or some other thread synchronization to ensure that between `Seek` and `Read`/`ReadByte` calls the position of the stream is not updated by another thread.

However, since this is a very hot path (AsmResolver's lazy eval heavily relies on random-access), I wonder about performance while moving to- and reading many individual bytes. This also makes me wonder about the added value of this, given that we already support MMIO backed data sources (which do allow for thread-safe random-access of a file). What is the main advantage that we get over MMIO?

Related: #115

ds5678:
In my case, I think performance would improve because I wouldn't need to copy the content to a byte array.

https://github.com/AssetRipper/AssetRipper/blob/e809ea198942c2e2fc882acc10daf3bc7ca37d17/Source/AssetRipper.Import/Structure/Assembly/Managers/BaseManager.cs#L95-L98

Uncontested locks are very lightweight for performance, so in synchronous code, any bottleneck would be during seeking or reading.

Since performance is a concern, I could replace the constructors with static create methods. This could allow creating specialized data sources based on the type of stream. In particular, memory streams seem like they would have a lot of potential for enabling random access.

Washi1337:
> Uncontested locks are very lightweight for performance, so in synchronous code, any bottleneck would be during seeking or reading.

True, I think we can leave potential performance hits to the consideration of the user of the API. People in general should use the normal (faster) random access versions of data sources where possible anyway.

PR #666, opened 2025-08-20
SignatureComparer ITypeDescriptor and TypeSignature changes

Windows10CE:
Namely fixes `Equals(typeSig, typeRef)`, but also generally makes an effort to use SimpleTypeEquals only when it makes sense. Leaves all of the hash code logic, since it already did this kind of deep searching via the Name/Namespace/Scope properties which are just forwarders to inner signatures.


Washi1337:
Thanks!

PR #668, opened 2025-08-24
Fix failure to resolve assembly with culture

Meivyn:
I'm assuming that this was the original intent, based on the comment that follows this block.


Washi1337:
Thanks for the catch!

PR #675, opened 2025-09-16
Add constructor overload for ExportedType

Sergio0694:
This PR adds a constructor overload for `ExportedType` to allow passing `Utf8String` values without round-tripping.


ds5678:
Since we're currently on a beta version and breaking changes are allowed, I think it would be better to just replace `string` with `Utf8String` rather than creating a new overload.

Washi1337:
Yea I concur with @ds5678, similar to #674 we should just get rid of the unnecessary overloads while we can. @Sergio0694 could you please change it? Then I will merge it shortly.

Sergio0694:
Done! 🙂

PR #676, opened 2025-09-19
Add `EventAttributes.None`

Sergio0694:
For consistency with `PropertyAttributes`. Also added a couple tests.


Washi1337:
Thanks!

PR #687, opened 2025-10-31
RuntimeAsync flag support

Windows10CE:
https://github.com/dotnet/runtime/issues/109632


Washi1337:
The changes in the MethodDef metadata verifier also warrants some unit tests. Could we add these tests without having to enable the experimental flags for the unit tests and/or test binaries?

PR #695, opened 2025-11-25
Optimize `BinaryStreamWriter`

Sergio0694:
This PR includes two optimizations for `BinaryStreamWriter`:
- On .NET 10, replaces the array with an inline array, to make that write optimization zero-alloc
- On .NS2.1/NC3.1 and above, it uses `stackalloc` or pooled arrays to write strings, to avoid allocating


Washi1337:
Thanks!

Note for future ref: This PR actually succeeded in CI pipeline. GitHub windows runners tests timed-out for whatever reason.

Issue #698, opened 2025-12-13
All `ITypeDescriptor`-s, `IMemberDescriptor`-s, etc. should implement `ISpanFormattable` and `IUtf8SpanFormattable`


Sergio0694:
### Problem Description

Right now, you have to allocate a lot of intermediate temporary string when reading their names (or calling `ToString()`):

<img width="1114" height="462" alt="Image" src="https://github.com/user-attachments/assets/4ac067dc-d0d3-4e2f-bfb4-f57d44effe87" />

This code calls `ToString()` on both, which then allocates a `string` every time.
Having `ISpanFormattable` would make all this zero-alloc.
Having `IUtf8SpanFormattable` would also allow this to skip all UTF8 transcoding, when needed.

### Proposal

All types implementing `ITypeDescriptor` and `IMemberDescriptor` should also implement `ISpanFormattable` and `IUtf8SpanFormattable`.

PR #699, opened 2025-12-13
Add `ISpanFormattable` and `IUtf8SpanFormattable` to `Utf8String`

Sergio0694:
Contributes to #698.

### DIFF
--- a/src/AsmResolver/Utf8String.cs
+++ b/src/AsmResolver/Utf8String.cs
@@ -374,13 +374,11 @@ namespace AsmResolver
             if (Value.TryCopyTo(destination))
             {
                 charsWritten = Value.Length;
-
                 return true;
 
             }
 
             charsWritten = 0;
-
             return false;
         }
 
@@ -398,12 +396,10 @@ namespace AsmResolver
             if (_data.AsSpan().TryCopyTo(utf8Destination))
             {
                 bytesWritten = _data.Length;
-
                 return true;
             }
 
             bytesWritten = 0;
-
             return false;
         }
 #endif


src/AsmResolver/Utf8String.cs
```diff
@@ -19,6 +19,10 @@ public sealed class Utf8String :
         IEquatable<byte[]>,
         IComparable<Utf8String>,
         IEnumerable<char>
+#if NET10_0_OR_GREATER
+        , ISpanFormattable
+        , IUtf8SpanFormattable
+#endif
```
Washi1337: `ISpanFormattable` is already available in .NET 6.0 ([msdn](https://learn.microsoft.com/en-us/dotnet/api/system.ispanformattable?view=net-10.0)) and `IUtf8SpanFormattable` in .NET 8 ([msdn](https://learn.microsoft.com/en-us/dotnet/api/system.iutf8spanformattable?view=net-9.0))

src/AsmResolver/Utf8String.cs
```diff
@@ -361,6 +365,45 @@ public int CompareTo(Utf8String? other) => other is not null
...
+            {
+                charsWritten = Value.Length;
+
+                return true;
+
+            }
+
+            charsWritten = 0;
+
+            return false;
```
Washi1337: Nit: Redundant extra blank lines.

src/AsmResolver/Utf8String.cs
Washi1337: We have to care because some users of AsmResolver also depend on dynamic features of the runtime (e.g., loading assemblies compiled for older runtimes with reflection alongside AsmResolver)

Sergio0694: Ah, gotcha. Fixed! 🙂

PR #705, opened 2026-01-25
Fix nonreproducibility in .text section

YoshiRulz:
Ran into this while trying to make [AssetRipper](https://github.com/AssetRipper/AssetRipper)'s output reproducible. No idea where in your codebase the problem is.

### DIFF
--- a/test/AsmResolver.DotNet.Tests/AsmResolver.DotNet.Tests.csproj
+++ b/test/AsmResolver.DotNet.Tests/AsmResolver.DotNet.Tests.csproj
@@ -52,6 +52,7 @@
             <Generator>ResXFileCodeGenerator</Generator>
             <LastGenOutput>Resources.Designer.cs</LastGenOutput>
         </EmbeddedResource>
+        <EmbeddedResource Include="Resources\mscorlib_LC.dll" />
     </ItemGroup>
 
     <ItemGroup>
--- a/test/AsmResolver.DotNet.Tests/ReproducibilityTest.cs
+++ b/test/AsmResolver.DotNet.Tests/ReproducibilityTest.cs
@@ -1,0 +1,44 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using AsmResolver.DotNet;
+using AsmResolver.DotNet.Builder;
+using AsmResolver.IO;
+using AsmResolver.PE;
+using AsmResolver.PE.Builder;
+using AsmResolver.PE.Exports;
+using AsmResolver.PE.Exports.Builder;
+using AsmResolver.PE.File;
+using AsmResolver.Tests.Runners;
+using Xunit;
+
+namespace AsmResolver.DotNet.Tests
+{
+    public class ReproducibilityTest : IClassFixture<TemporaryDirectoryFixture>
+    {
+        [Fact]
+        public void TextSection()
+        {
+            byte[] CreateImageAndHashTextSection()
+            {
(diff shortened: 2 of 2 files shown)


Washi1337:
Please read [CONTRIBUTING.md](https://github.com/Washi1337/AsmResolver/blob/master/CONTRIBUTING.md), base your branch off of `development` and adhere to coding style.

There is also no need to fully embed `mscorlib.dll` into the tests project, if it can be fetched from the disk.

> No idea where in your codebase the problem is.

Then this should really be an issue where we could discuss the problem. Thus, closing this PR. Feel free to create a PR if you have a solution implemented.

YoshiRulz:
To be clear: I intend to track down and fix the bug. In opening a *draft* PR, my hope was that you might be able to tell at a glance where the problem might be.
I can't change the PR to target `development` since you've closed it. I've made an attempt at addressing your other two nitpicks locally, but IIRC pushing to a branch that's tied to a closed PR breaks it.

Windows10CE:
It has been pointed out on Discord that the issue is likely not setting ModuleDefinition.Mvid, as by default in a newly created module it gets a newly generated GUID. Changing your test to add `module.Mvid = Guid.Empty;` does indeed make it pass.

YoshiRulz:
Thank you!
[I copied](https://github.com/YoshiRulz/AsmResolver/commit/e720e8bfbd813092ee37b61743932e67b44fb9c9) what I did [for AssetRipper](https://github.com/AssetRipper/AssetRipper/pull/2057/changes/91e41f1da38023b68964854a734a53e994c5d8e8) to make random GUIDs deterministic, and changed the test to compare against a hardcoded value. Though I guess the test isn't really needed anymore.

Washi1337:
>  I can't change the PR to target development since you've closed it. I've made an attempt at addressing your other two nitpicks locally, but IIRC pushing to a branch that's tied to a closed PR breaks it.

It's possible to create a new PR from the same branch or a different branch.

> Thank you!
> [I copied](https://github.com/YoshiRulz/AsmResolver/commit/e720e8bfbd813092ee37b61743932e67b44fb9c9) what I did [for AssetRipper](https://github.com/AssetRipper/AssetRipper/pull/2057/changes/91e41f1da38023b68964854a734a53e994c5d8e8) to make random GUIDs deterministic, and changed the test to compare against a hardcoded value. Though I guess the test isn't really needed anymore.

Looking at your commit on your fork, I am very hesitant to let AsmResolver's behavior depend on global environment variables for what could also be just an input parameter for a PE image builder.

A library's behavior should be as little influenced by global state as possible. The more global state there is, the more we lock consumers into a certain paradigm (e.g., the use of an env var) and the more unpredictable the library gets (e.g., different consumers may assume different mismatched behaviors). With a parameter, a consumer can independently decide how it wants to use the builder or not.

YoshiRulz:
Well `ModuleDefinition.Mvid` is settable. So should it be up to the caller to set that on init, every time?

Washi1337:
> Well `ModuleDefinition.Mvid` is settable. So should it be up to the caller to set that on init, every time?

Yes, a new blank `ModuleDefinition` should have a random MVID. It is left intentionally settable such that for scenarios where a "deterministic" mvid is needed, you can use your own deterministic algorithm for it. This is by design.

YoshiRulz:
Alright then. It turns out that that was all I needed to get reproducible results in AssetRipper, so unless you think `MemberDiscoverer` or `JSONObject` need fixing, I won't bother opening a new PR. Sorry for wasting your time.

PR #710, opened 2026-02-09
Add WindowsRuntime attribute and property

Sergio0694:
Introduce a WindowsRuntime flag to the TypeAttributes enum and add an IsWindowsRuntime boolean property on TypeDefinition. The property reads and updates the Attributes bitmask using TypeAttributes.WindowsRuntime, enabling types to be marked and queried as Windows Runtime types.


Washi1337:
Thanks!

PR #711, opened 2026-02-10
Add thread storage and trampoline PDB types

dongle-the-gadget:
**Additional consideration:** The Trampoline type has extra data outside of the header. Not sure how I'm supposed to present that.


Washi1337:
Thanks for your contribution!

Do we have any sample PDB file that we can create to include in unit tests? This way, we can verify not only that the current implementation works, but also that it remains working in case we make changes in the future.

dongle-the-gadget:
Do you mean for these scenarios or for all scenarios in general? For a Hello World file with printf I could get Trampolines but not thread storage.

Washi1337:
> Do you mean for these scenarios or for all scenarios in general? For a Hello World file with printf I could get Trampolines but not thread storage.

Ideally, we would have a PDB that contains the records you're adding support for (including at least one record of thread storage), so that we can programmatically test against it.

I was able to get a gthread32 symbol using a `thread_local` global variable:

```cpp
#include <iostream>
#include <thread>

thread_local int foo;

int main()
{
    foo = 1337;
    printf("Foo (main thread): %d\n", foo);
    std::thread([] {
        printf("Foo (other thread): %d\n", foo); 
    }).join();
    return 0;
}
```

image

As for what is to be tested specifically per record: There's no need to fully capture all possibilities, such as all different kinds of type records that can be assigned to `VariableType` in thread storage (this is already covered by other tests specifically related to type records alone). We just want to make sure that the fields stored in each record are generally speaking read properly.

dongle-the-gadget:
I've added tests for thread locals. Not sure how to approach trampolines yet. Should we try to i.e. retrieve a known-good list from [pdb-rs](https://github.com/microsoft/pdb-rs) and compare AsmResolver's results?

Washi1337:
> I've added tests for thread locals. Not sure how to approach trampolines yet. Should we try to i.e. retrieve a known-good list from [pdb-rs](https://github.com/microsoft/pdb-rs) and compare AsmResolver's results?

Yes, any data pulled from _some_ ground-truth source is fine.

Also, please if possible, put relevant tests per records in their own separate files (eg. see existing tests like [ThunkSymbolTest.cs](https://github.com/Washi1337/AsmResolver/blob/master/test/AsmResolver.Symbols.Pdb.Tests/Records/ThunkSymbolTest.cs) for example)

dongle-the-gadget:
@Washi1337 I've moved the thread storage tests and added the trampoline tests (fixed a bug in the trampoline reader while I'm at it 😅)

Washi1337:
Thanks for your contribution!

PR #712, opened 2026-02-11
Add 'CreateConstructor' overload taking just 'CorLibTypeFactory'

Sergio0694:
Especially with the upcoming type resolution changes, it won't be needed to always have a `ModuleDefinition` being passed around. As part of this, we're trying to actually only ever pass it where we're adding types to it, otherwise we only pass the set of references being used. This also allows us to more clearly separate:
- Factory methods that produce some stuff
- Builder methods that actually assemble/add things to a target module

The current `CreateConstructor` method is annoying because it forces you to pass a `ModuleDefinition` only so it can use its type factory. This PR just adds an overload that takes it directly, so you don't need a module on the caller side at all anymore.


Washi1337:
Thanks!

PR #715, opened 2026-02-20
Improved leaf record errors

ds5678:
See previous version: #360


Washi1337:
Not sure what is going on here, why are all tests failing with this change?

Also, from a design perspective, I don't think it is a good idea to make a public `GetLeafRecord` dependent on an error listener. That goes against the design of other APIs (e.g.,`ModuleDefinition.LookupMember` also doesn't depend on the error listener of the underlying reader).

Besides, if we want improved error messages, why wouldn't we just leave the error handling in the respective `GetX` initializer methods? Wouldn't that allow for much more fine-grained diagnostics as it is now?

ds5678:
> Not sure what is going on here, why are all tests failing with this change?

I didn't look into it.

> Also, from a design perspective, I don't think it is a good idea to make a public `GetLeafRecord` dependent on an error listener. That goes against the design of other APIs (e.g.,`ModuleDefinition.LookupMember` also doesn't depend on the error listener of the underlying reader).

Understood

> Besides, if we want improved error messages, why wouldn't we just leave the error handling in the respective `GetX` initializer methods? Wouldn't that allow for much more fine-grained diagnostics as it is now?

I will abandon this pull request for now. If I come to this pull request, I think I can address your concerns with a source generator for those methods. It seems like an ideal solution to the problem.

PR #717, opened 2026-02-24
Avoid unnecessary string allocations in CorLibTypeFactory.FromType

wondercrash:
Before:
image
After:
image

Some type signatures (e.g. SzArray) aren't corlib types, and accessing their `Name` property can allocate a string. This PR ensures a type can actually be a corlib type before matching its name.

I believe the logic is correct for this. I would welcome a second set of eyes though.

### DIFF
--- a/src/AsmResolver.DotNet/Signatures/CorLibTypeFactory.cs
+++ b/src/AsmResolver.DotNet/Signatures/CorLibTypeFactory.cs
@@ -153,9 +153,17 @@ namespace AsmResolver.DotNet.Signatures
         /// <returns>The corlib type, or <c>null</c> if none was found.</returns>
         public CorLibTypeSignature? FromType(ITypeDescriptor type)
         {
-            return type is CorLibTypeSignature typeSig
-                ? FromElementType(typeSig.ElementType)
-                : FromName(type.Namespace, type.Name);
+            if (type is CorLibTypeSignature typeSig)
+                return FromElementType(typeSig.ElementType);
+
+            // Corlib types are always encoded as TypeDefOrRefSignature (ValueType/Class).
+            // Other TypeSignatures are structural types (arrays, pointers, byrefs, etc.) that can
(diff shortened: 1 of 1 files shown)


Windows10CE:
The comment is *technically* wrong, other signatures can match, specifically CustomMod, Pinned, or Boxed. Whether it is worth checking those as well or if that was actually *incorrect* behavior and not matching those is correct is another question.

Washi1337:
> The comment is technically wrong, other signatures can match, specifically CustomMod, Pinned, or Boxed. Whether it is worth checking those as well or if that was actually incorrect behavior and not matching those is correct is another question.

Strictly speaking, the original implementation would've already ruled out CustomMods, because CustomModifierTypeSignature overrides the Name property by including the modreq/modopt syntax. If we want to keep consistent with previous implementations, we only have to check for Boxed and Pinned (and _technically_ any number of repetitions of those in sequence).

That being said, I think it is fine to not include them, and only limit ourselves to TypeDefs and TypeRefs. Technically, all others are a different type signature with different runtime behavior afaik.

wondercrash:
> The comment is _technically_ wrong, other signatures can match, specifically CustomMod, Pinned, or Boxed. Whether it is worth checking those as well or if that was actually _incorrect_ behavior and not matching those is correct is another question.

Good point. I think it's more correct to not match them since those are modifier types. They could change the memory layout (e.g. pinned int32 is 8 bytes, but a normal int is 4 bytes) so it might be good to preserve them. It also didn't _explicitly_ unwrap the modifier types into the corlib types before. 

My only worry with not matching them is the memory layout thing I mentioned. [Echo in particular](https://github.com/Washi1337/Echo/blob/57b9c22339cf8886aab30fec2b039b1cd6084da1/src/Platforms/Echo.Platforms.AsmResolver/Emulation/ValueFactory.cs#L466-L474). But that is just something that would have to be fixed per project.

Washi1337:
Thanks!

PR #719, opened 2026-02-24
Further optimize untouched SerializedCustomAttributeSignature

js6pak:


Washi1337:
Is this still necessary with this already in place:

https://github.com/Washi1337/AsmResolver/blob/a6d8b149879bc5ef6fb9e719e6185d863f5c436c/src/AsmResolver.DotNet/Signatures/SerializedCustomAttributeSignature.cs#L72-L75

What do we gain from this?

js6pak:
> What do we gain from this?

After benchmarking again, turns out basically nothing anymore... This is a patch from forever ago and it seems like other changes made this unnecessary.

PR #736, opened 2026-04-26
Docs: Add a small note about native methods

Yoticc:
While using AsmResolver, I ran into an issue where the assembly containing my created native methods (following the guide) would not load. It took me a long time to figure out the reason, which I have now documented in the commit note.
Even though the MSIL specification does not explicitly state that native methods must be defined in <Module>, the CoreCLR implementation includes code in coreclr\vm\methodtablebuilder.cpp [MethodTableBuilder::ValidateMethods] that checks for IsGlobalClass(). Because of this, I think this note is worth existing

### DIFF
--- a/src/AsmResolver.DotNet/Signatures/TypeSignature.cs
+++ b/src/AsmResolver.DotNet/Signatures/TypeSignature.cs
@@ -555,14 +555,12 @@ namespace AsmResolver.DotNet.Signatures
 
         /// <summary>
         /// Constructs a new boxed type signature with the provided type descriptor as element type.
-        /// as element type.
         /// </summary>
         /// <returns>The constructed boxed type signature.</returns>
         public BoxedTypeSignature MakeBoxedType() => new(this);
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


docs/guides/dotnet/unmanaged-method-bodies.md
```diff
@@ -22,6 +22,12 @@ your mixed mode application, and might throw runtime or image format
 exceptions, even if everything else conforms to the right format. This
 section will go over these requirements briefly.
 
+> [!NOTE]
+> There are restrictions on the location of native methods in the 
+> CLR core implementation: they can only be located in the global <Module> class.
```
Washi1337: Use backticks:
```
`<Module>`
```

Washi1337 (commented):
Small comment but other than that lgtm. Also, please base your PR on the `development` branch instead of `master`.

Yoticc:
> Small comment but other than that lgtm. Also, please base your PR on the `development` branch instead of `master`.

oops, my bad. i hope github's branch change works as expected and i will not have to create a new pull request

PR #737, opened 2026-04-26
Docs: add a note about native method restrictions

Yoticc:
While using AsmResolver, I ran into an issue where the assembly containing my created native methods (following the guide) would not load. It took me a long time to figure out the reason, which I have now documented in the commit note.
Even though the MSIL specification does not explicitly state that native methods must be defined in , the CoreCLR implementation includes code in coreclr\vm\methodtablebuilder.cpp [MethodTableBuilder::ValidateMethods] that checks for IsGlobalClass(). Because of this, I think this note is worth existing


Washi1337:
Thanks!

GitHub actions fluked here for some reason (once again...).

PR #740, opened 2026-05-07
Add parentheses to not..or pattern

Arthri:
I'm failing to build the master branch with the .NET 10.0.203 SDK because of the CS9336 error reported here. I assume from the exception message below it, the condition is missing parentheses after `not` which I have added


Washi1337:
Duplicate of #732. Development is done on the `development` branch and synced with nightly builds, `master` is in sync with public nuget/releases. See also CONTRIBUTING.md

PR #742, opened 2026-05-07
Add IsAssignableFrom

Arthri:
Added two `IsAssignableFrom` overloads that call the two `IsAssignableTo` methods with the same signature with the `this` and `other` arguments swapped.

I also renamed `IsAssignableToForwardedType` to `IsAssignableForwardedType` and added a test data generator with the four permutations per type-forwarded type pair so the resulting test with both `IsAssignableTo` and `IsAssignableFrom` is shorter


Washi1337:
Thanks!

GitHub actions fluked again. We may have to invest some time in CI/CD pipeline stability again.

Issue #66, opened 2020-06-06
Support for reading and writing Win32 resources


MaxXor:
**Summary:**
Currently, Win32 resources can only be read from `IReadableSegment`, which can be used to create an `IBinaryStreamReader` and read the resource in raw bytes. Updating and interpreting of resources, such as [VERSIONINFO](https://docs.microsoft.com/en-us/windows/win32/menurc/versioninfo-resource) resources, is left to the user and cumbersome to parse. Native support for writing/reading such resources with parsed information would make this easier.

A look at how it's done in Vestris.ResourceLib for [VERSIONINFO](https://github.com/resourcelib/resourcelib/wiki/Version-(RT_VERSION)) resources might be helpful.

Washi1337:
I'd say this is perfect for a separate package (not everyone might need this feature). I added the `AsmResolver.PE.Win32Resources` project to the solution.

I'd also say all resource types should be assigned their own namespace in the  For example, I placed the version info resource into the `AsmResolver.PE.Win32Resources.Version` namespace.

Do you have any suggestions for the public API to make it easier to get to the interpreted version of the resources?

laomms:
Is it possible make it support update string table.

PR #745, opened 2026-05-19
Add RT_STRING (StringTable) resource parser

JPaja:
Adds a high-level reader/writer for RT_STRING (String Table) Win32 resources. Partially addresses #66.


src/AsmResolver.PE.Win32Resources/StringResource/StringTableResource.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        foreach (var block in stringDirectory.Entries.OfType<ResourceDirectory>())
+        {
+            uint blockId = block.Id;
+
+            foreach (var language in block.Entries.OfType<ResourceData>())
+            {
+                var reader = language.CreateReader();
+                ReadBlock(result, blockId, ref reader);
+            }
```
Washi1337, line 39: If the resources contain multiple string tables for different languages, I believe this will only preserve the last language specified.

src/AsmResolver.PE.Win32Resources/StringResource/StringTableResource.cs
```diff
@@ -0,0 +1,111 @@
...
+                continue;
+
+            uint byteCount = (uint)(length * 2);
+            if (!reader.CanRead(byteCount))
+                break;
+
+            byte[] data = new byte[byteCount];
+            reader.ReadBytes(data, 0, (int)byteCount);
+            string value = Encoding.Unicode.GetString(data, 0, data.Length);
+            resource.Strings[baseId + (uint)i] = value;
```
Washi1337, line 62: Should we use `Utf8String.CreateUnsafe` instead (i.e., to preserve potential invalid encodings of UTF8)?

src/AsmResolver.PE.Win32Resources/StringResource/StringTableResource.cs
```diff
@@ -0,0 +1,111 @@
...
+                }
+                else
+                {
+                    writer.WriteUInt16((ushort)s.Length);
+                    writer.WriteBytes(Encoding.Unicode.GetBytes(s));
+                }
+            }
+
+            var blockDir = new ResourceDirectory(kvp.Key);
+            blockDir.Entries.Add(new ResourceData(0, new DataSegment(stream.ToArray())));
```
Washi1337, line 105: This forces everything to be put in as a language-neutral data entry. We should preserve all language-specific entries.

Washi1337 (changes requested):
I believe the current approach does not support multiple languages. I think it would make sense to extract the `IDictionary<uint, string>` into a separate class that is representative of a single language.

Maybe something like the following:

```csharp
class StringTableResource
{
    public StringTable GetTable(int lcid);
    public IEnumerable<StringTable> GetNonEmptyTables(); 
}

class StringTable
{
    public IDictionary<uint, Utf8String> Entries { get; }
}
```

In this setup, `StringTable` would be a "sub-view" on all the string entries for a specific language, and `StringTableResource::GetTable` gets or lazily creates the table for the provided language. `StringTableResource::InsertIntoDirectory` should then go over all non-empty string tables and reconstruct the raw resource directory tree.

PR #752, opened 2026-06-08
Use params collection parameters for generic arguments

ds5678:
This is a non-breaking change.


Washi1337:
Params keywords were removed deliberately to stay consistent with the changes of the `MethodSignature` factory methods. C# collection expressions are the way forward.

PR #753, opened 2026-06-08
Change IsTrimmable to IsAotCompatible

### LINKED ISSUE #751: Possible NativeAOT compatibility?
### Description

Hello!
I have doing some preemptive research into libraries that allow reading managed .NET assemblies and came across AsmResolver. I would like to use it in a project that is planned to be deployed via NativeAOT.
However, when adding it as a package reference and compiling a test project, I get a warning that it might not be AOT compatible:
*(I don't know why it says "built with \`true\`, but that's funny)*

    0>CSC: Warning IL3058 : Referenced assembly 'AsmResolver' is not built with `true` and may not be compatible with AOT.
    0>CSC: Warning IL3058 : Referenced assembly 'AsmResolver.DotNet' is not built with `true` and may not be compatible with AOT.
    0>CSC: Warning IL3058 : Referenced assembly 'AsmResolver.PE' is not built with `true` and may not be compatible with AOT.
    0>CSC: Warning IL3058 : Referenced assembly 'AsmResolver.PE.File' is not built with `true` and may not be compatible with AOT.

I did find that there is a NativeAOT project that uses AsmResolver in [`showcase.md`](https://github.com/Washi1337/AsmResolver/blob/47741da8e2562465081cde59c066b0db78bcdd46/docs/guides/showcase.md?plain=1#L10), though I couldn't find any clear statement that AsmResolver itself supports NativeAOT.

So, does AsmResolver support NativeAOT? Are there any known considerations and caveats that I should be aware of when using it in a NativeAOT setting?

Thank you for your time!

ds5678:
IsAotCompatible is a stronger version of IsTrimmable.

I checked to see if any code needed additional attribute annotations due to the higher strictness, but I encountered no warnings while building.

Resolves #751

### DIFF
--- a/src/AsmResolver.DotNet.Dynamic/AsmResolver.DotNet.Dynamic.csproj
+++ b/src/AsmResolver.DotNet.Dynamic/AsmResolver.DotNet.Dynamic.csproj
@@ -4,7 +4,7 @@
     <Title>AsmResolver.DotNet.Dynamic</Title>
     <Description>Dynamic method support for the AsmResolver executable file inspection toolsuite.</Description>
     <PackageTags>exe pe directories imports exports resources dotnet cil inspection manipulation assembly disassembly dynamic</PackageTags>
-    <IsTrimmable>false</IsTrimmable>
+    <IsAotCompatible>false</IsAotCompatible>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


Washi1337:
Thanks!

PR #756, opened 2026-06-27
BUGFIX: Correct strong-name RSA key blob encoding

Summpot:
## Problem
Strong-name key blobs use two different byte-order conventions at the same API boundary:

- .NET `RSAParameters` stores RSA integers such as `Modulus`, `P`, `Q`, `DP`, `DQ`, `InverseQ`, and `D` as big-endian byte arrays.
- CryptoAPI `PUBLICKEYBLOB` and `PRIVATEKEYBLOB` store those same RSA integers in little-endian order on disk.
- The CryptoAPI `pubexp` field is different again: it is a 32-bit integer field, so it is serialized as a little-endian `uint`, while `RSAParameters.Exponent` is still a big-endian, no-leading-zero byte array.

The previous implementation mixed these representations. In particular, a strong-name key constructed from `RSAParameters` could be serialized as a CryptoAPI blob with big-endian RSA integer bytes, even though consumers expect little-endian blob integers. Reading public-key blobs also left the modulus in the on-disk byte order, so converting it back to `RSAParameters` could produce parameters that `RSA.ImportParameters` rejects or interprets incorrectly.

The exponent issue was easy to miss because the common exponent `65537` is encoded by `RSAParameters` as `[01 00 01]`, which reads the same forwards and backwards. Non-palindromic exponent byte sequences expose the wrong conversion.

`StrongNamePublicKey.GetPhysicalSize()` also divided the modulus byte length by 8, causing the reported serialized size to be smaller than the number of bytes actually written.

## Fix
- Use `RSAParameters`' big-endian byte order as the internal representation for strong-name RSA integer fields.
- Reverse RSA integer byte arrays only when reading from or writing to CryptoAPI key blobs.
- Convert `RSAParameters.Exponent` as a big-endian integer and serialize `pubexp` as the CryptoAPI little-endian `uint32` field.
- Fix `StrongNamePublicKey.GetPhysicalSize()` to include the full modulus byte length.
- Add regression tests covering CryptoAPI blob layout, non-palindromic exponent conversion, RSA import/signing round-trips, and physical size consistency.

## Tests
- `dotnet format whitespace --verify-no-changes --include src/AsmResolver.PE/DotNet/StrongName/StrongNamePublicKey.cs src/AsmResolver.PE/DotNet/StrongName/StrongNamePrivateKey.cs test/AsmResolver.PE.Tests/DotNet/StrongName/StrongNamePrivateKeyTest.cs`
- `dotnet test test/AsmResolver.PE.Tests/AsmResolver.PE.Tests.csproj`

### DIFF
--- a/src/AsmResolver.PE/DotNet/StrongName/StrongNamePrivateKey.cs
+++ b/src/AsmResolver.PE/DotNet/StrongName/StrongNamePrivateKey.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Security.Cryptography;
 using AsmResolver.IO;
 
@@ -39,7 +38,7 @@ namespace AsmResolver.PE.DotNet.StrongName
             ReadBlobHeader(ref reader, StrongNameKeyStructureType.PrivateKeyBlob, 2, SignatureAlgorithm.RsaSign);
 
             // Read RSAPUBKEY
-            if ((RsaPublicKeyMagic) reader.ReadUInt32() != RsaPublicKeyMagic.Rsa2)
+            if ((RsaPublicKeyMagic)reader.ReadUInt32() != RsaPublicKeyMagic.Rsa2)
                 throw new FormatException("Input stream does not contain a valid RSA private key header magic.");
 
             uint bitLength = reader.ReadUInt32();
@@ -49,9 +48,12 @@ namespace AsmResolver.PE.DotNet.StrongName
                 PublicExponent = reader.ReadUInt32(),
             };
 
+            // CryptoAPI PRIVATEKEYBLOB stores every RSA integer (Modulus, P, Q,
+            // DP, DQ, InverseQ, D) in little-endian. The in-memory representation
+            // here is big-endian (matching StrongNamePublicKey's storage and the
+            // big-endian RSAParameters that ToRsaParameters returns), so each
+            // integer is read as raw little-endian bytes and then reversed.
             reader.ReadBytes(result.Modulus, 0, result.Modulus.Length);
-
-            // Read private data.
             reader.ReadBytes(result.P, 0, result.P.Length);
             reader.ReadBytes(result.Q, 0, result.Q.Length);
             reader.ReadBytes(result.DP, 0, result.DP.Length);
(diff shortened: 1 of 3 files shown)


src/AsmResolver.PE/DotNet/StrongName/StrongNamePrivateKey.cs
```diff
@@ -90,16 +92,20 @@ public StrongNamePrivateKey(uint bitLength)
         /// <summary>
         /// Imports a public/private key pair from an instance of <see cref="RSAParameters"/>.
         /// </summary>
-        /// <param name="parameters">The RSA parameters to import.</param>
+        /// <param name="parameters">The RSA parameters to import. All integers
+        /// are expected in the big-endian byte order <see cref="RSA"/> uses and
+        /// are stored verbatim (big-endian) so <see cref="ToRsaParameters"/>
+        /// returns them unchanged.</param>
```
Washi1337: The big-endian mention is fine, but let's leave out internal implementation details like this.

src/AsmResolver.PE/DotNet/StrongName/StrongNamePrivateKey.cs
```diff
@@ -171,23 +177,23 @@ public byte[] PrivateExponent
         }
 
         /// <inheritdoc />
+        /// <remarks>
+        /// All integers are stored internally big-endian (the byte order
+        /// <see cref="RSA.ImportParameters"/> expects), so they are returned
+        /// verbatim. The on-disk CryptoAPI little-endian layout is handled by
+        /// reversing in <see cref="Write"/> and <see cref="FromReader"/>.
+        /// </remarks>
         public override RSAParameters ToRsaParameters()
```
Washi1337: These implementation details are unnecessary to expose as xmldoc. Users should not have to care about this.

src/AsmResolver.PE/DotNet/StrongName/StrongNamePrivateKey.cs
```diff
@@ -215,24 +221,31 @@ public override uint GetPhysicalSize()
...
-            uint exponent = 0;
-            for (int i = 0; i < Math.Min(sizeof(uint), parameters.Exponent.Length); i++)
-                exponent |= (uint) (parameters.Exponent[i] << (8 * i));
-            return exponent;
+            // RSAParameters.Exponent is big-endian (no leading zeros). Read it
+            // as big-endian into the uint. (The previous little-endian read
+            // happened to work for the palindromic 65537 = [01 00 01] but produced
+            // the wrong uint for any other exponent, e.g. [01 00 02] would become
+            // 0x020001 instead of 0x010002.)
+            return BigEndianBytesToUInt(parameters.Exponent);
```
Washi1337: We don't need comments for previous mistakes.

src/AsmResolver.PE/DotNet/StrongName/StrongNamePublicKey.cs
```diff
@@ -41,24 +41,85 @@ public static StrongNamePublicKey FromReader(ref BinaryStreamReader reader)
...
+                return new byte[] { 0 };
+
+            int length = value <= 0xFF
+                ? 1
+                : value <= 0xFFFF
+                    ? 2
+                    : value <= 0xFFFFFF
+                        ? 3
+                        : 4;
+            var result = new byte[length];
```
Washi1337: Can we do something about this staircase ternary to improve readability? Maybe a switch expression?
```csharp
int length = value switch
{
   ...
};
```

src/AsmResolver.PE/DotNet/StrongName/StrongNamePublicKey.cs
```diff
@@ -134,30 +194,35 @@ public byte[] CreatePublicKeyBlob(AssemblyHashAlgorithm hashAlgorithm)
...
+            writer.WriteBytes(Reverse(Modulus));
             return tempStream.ToArray();
         }
 
         /// <summary>
         /// Translates the strong name parameters to an instance of <see cref="RSAParameters"/>.
         /// </summary>
-        /// <returns>The converted RSA parameters.</returns>
+        /// <returns>The converted RSA parameters, with all integers in the
+        /// big-endian byte order <see cref="RSA.ImportParameters"/> expects.</returns>
```
Washi1337: I don't think we need this clarification on internals for the user to read.

Washi1337 (changes requested):
Thanks for the catch. I am not sure how signing ever worked before this change (:

Few comments but otherwise LGTM: We do not need to over-explain internal workings of these classes to the user in xmldoc. This way we do not lock ourselves to this implementation only in the future, and also it's just going to be confusing for users in general I feel.

Washi1337:
Thanks for looking into it!

PR #760, opened 2026-07-26
AsmResolver.Benchmark improvements

3fbaea00:
Updated BenchmarkDotNet version.
Added hardware counters to benchmarks.
Switched to a newer benchmark execution pipeline.


3fbaea00:
raaaah, GITUHB

PR #762, opened 2026-07-26
[DeUnsafication & ByRefication] Simplified, moved away from unsafe manual implementation

3fbaea00:
For .netstandard2.0 and lower:
\- 500 bytes native size
```
| Method   | Mean     | Error    | StdDev   |
|---------:|---------:)---------:/---------:o
| Old      | 40.77 ns | 0.235 ns | 0.219 ns |
| New      | 21.26 ns | 0.281 ns | 0.249 ns |
```


3fbaea00:
Lost relevance because of #773

PR #765, opened 2026-07-27
Add note about verify argument defaul value in xml-documentation of MethodDefinition constructor

3fbaea00:


Washi1337:
The parameter does not have a default value, so the extra comment seems wrong?

3fbaea00:
> The parameter does not have a default value, so the extra comment seems wrong?

there is another constructor without this argument and it is confusing, since it is not clear, which version of verify will be used when this argument is omitted

Washi1337:
I don't see a reason to add it to this constructor. A different constructor means reading different xml documentation.

PR #766, opened 2026-07-29
Add missed machine type enums

3fbaea00:
some of these added architectures are used by r2r, so they have a place here


Washi1337:
Is there a source for this? (for reference).

3fbaea00:
i will move it to #769

PR #767, opened 2026-07-29
Restrict the publicity of the method semantics

3fbaea00:
public access to this property setter bypasses the correct logic set by semantics collection

### DIFF
--- a/src/AsmResolver.DotNet/MethodDefinition.cs
+++ b/src/AsmResolver.DotNet/MethodDefinition.cs
@@ -746,7 +746,7 @@ namespace AsmResolver.DotNet
         public partial MethodSemantics? Semantics
         {
             get;
-            set;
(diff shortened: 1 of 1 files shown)


Washi1337:
Thanks for the catch. I think this can even be made private, no? As far as I can see (in GitHub search), we never explicitly assign the Semantics property anywhere but by the source generators.

3fbaea00:
> Thanks for the catch. I think this can even be made private, no? As far as I can see (in GitHub search), we never explicitly assign the Semantics property anywhere but by the source generators.

i did not really understand the idea. make the setter private? how will MethodSemanticsCollection be able to assign it then? it accesses the method semantics directly, modifying them

Washi1337:
My bad, missed this. GitHub code search is awful.

PR #768, opened 2026-07-29
To ElementType enum add more definitions related to r2r and SerializationType.

3fbaea00:
Also add minimal xml-doc comments, they are quite modest.

For zapsig, an alternative could be to define:
```cs
/// <summary>
/// Extends <see cref="ElementType"/> by runtime-specific definitions required for ReadyToRun and Multi-core JIT recording/playback.
/// </summary>
public enum RuntimeElementType : byte
```
and in it redefining all ElementType members, as well as additional runtime-specific signatures.
But since the current implementation contains definitions for the Serialization Type enum, i felt that updating the ElementType enum would be the right way.


3fbaea00:
moved to #769

PR #769, opened 2026-07-29
ReadyToRun support

3fbaea00:
Instead of a bunch of pull requests, i will gradually upload the progress here


3fbaea00:
New monomod version ^^^ is required at least for BitOperations

3fbaea00:
My current vision is a separate nuget package AsmResolver.DotNet.ReadyToRun, with extensions to the dotnet part, containing all managed structures, as well as several checks for lazy initialization of all data into managed structures, and write checks to determine if the section needs to be written back to the output file or if the bytes can simply be copied from the original file. also, there could be an AutomaticallyRebuildR2RSection method here to evaluate which original R2R structures can have some entities removed due to their disappearance from the dotnet structures, e.g., method removal, as well as attribute changes, e.g., a method without an R2R native body was added and the partial flag needs to be added. In Asmresolver.PE, only native data related by relative links, enumerations, and sections are stored. I would rather represent them as a byte array and methods for reading them. For example, a large bitwise structure contains the necessary number of bytes for writing directly from the section, as well as additional shifts (in case the structure has optional fields)

PR #771, opened 2026-07-29
Mark `TypeSignature::Name` not null

ds5678:
I split this into 2 commits in case the second commit is considered too big of a breaking change.


Washi1337:
Thanks!

PR #773, opened 2026-07-31
Push monomod backports

3fbaea00:
Disable generate NuGet package on build to speed up building process.

AllowUnsafeBlocks/GeneratePackageOnBuild to Directory.Build.props.

Remove IsExternalInit and Nullable NuGet dependencies. Add global MonoMod.Backport 2.0.0 prerelease.5 NuGet dependency.
Due to the prerelease version of the dependency, it requires NoWarn=NU5104 to pack a NuGet package.

Through the backports version update, span and other new versions elements became available, i tried to remove all pieces of explicit .netstandard support that i could.
Visual studio ate the spaces in casts, it will be necessary to make another commit with a global solution to this problem for visual studio and another one bug


Washi1337:
I'm not sure I understand the proposed changes.

- This has a lot of unnecessary code style changes all over the place. Please address them to minimize the diff.
- Setting `GeneratePackageOnBuild` to `false` would break the CI pipeline on dev/master branch merge, since it depends on the nuget packages to be present in the build artifacts. See also: https://github.com/Washi1337/AsmResolver/blob/5aa0629348c523458780846b54a51f23ab8afced/.github/workflows/test-and-publish.yml#L253-L266
- I am also not sure why we need to set `AllowUnsafeBlocks` to `true` globally in `src`, or always add `MonoMod.Backports` even for build targets that don't need it (i.e. modern .NET).
- `IsExternalInit` and `Nullable` are source generators that conditionally adds the required attributes as embedded attributes in the assembly. As far as I know, MonoMod.Backports does not do this?
- Bumping MonoMod.Backports a major version is a breaking change. Also, I am not sure I am comfortable with using a prerelease package in production.

3fbaea00:
> * Setting `GeneratePackageOnBuild` to `false` would break the CI pipeline on dev/master branch merge, since it depends on the nuget packages to be present in the build artifacts.

Does not that mean it can be added an extra job for "dotnet pack -o ./{MSBuildProjectName}/artifacts" to allow GeneratePackageOnBuild = false?

> * I am also not sure why we need to set `AllowUnsafeBlocks` to `true` globally in `src`, or always add `MonoMod.Backports` even for build targets that don't need it (i.e. modern .NET).

Why not consider AllowUnsafeBlocks in every project? I suspect the reason might be future unsafe restrictions in dotnet?

> * `IsExternalInit` and `Nullable` are source generators that conditionally adds the required attributes as embedded attributes in the assembly. As far as I know, MonoMod.Backports does not do this?

These types are located in the MonoMod.Backports.dll assembly and not optionaly embedded in the target assembly.

> * Bumping MonoMod.Backports a major version is a breaking change. Also, I am not sure I am comfortable with using a prerelease package in production.

Yes, at least, this triggers warnings with pack (which is why NoWarn=NU5104 is required).
As far as I understand, MonoMod.Backports is not ready for release. But it is best to ask @nike4613 about it.

> * This has a lot of unnecessary code style changes all over the place. Please address them to minimize the diff.

I apologize. As I said, Visual Studio did it. Once the points above are resolved, I will mark each change.

Washi1337:
> Does not that mean it can be added an extra job for "dotnet pack -o ./{MSBuildProjectName}/artifacts" to allow GeneratePackageOnBuild = false?

Yes, we would need to add something like this to the final nuget step of the CI/CD pipeline.

> Yes, at least, this triggers warnings with pack (which is why NoWarn=NU5104 is required).

Regardless of prerelease status, a major version bump of a dependency would require a major version bump for AsmResolver as well. I am not necessarily against using only MonoMod.Backports for this and getting rid of the other attributes, but we would have to wait until v7 comes around.

> Why not consider AllowUnsafeBlocks in every project? I suspect the reason might be future unsafe restrictions in dotnet?

Because I don't think we need it? If we can avoid unsafe then we probably should. The use of unsafe should be a deliberate choice, not the default.

PR #774, opened 2026-07-31
Fix editor config for visual studio

3fbaea00:
Fixed the issue where visual studio does not add a space after a cast.
Example: (int)MyVariable instead of the correct (int) MyVariable

Fixed the issue with adding a new line for slnx files that caused a soft lock in VS 18

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,7 +13,6 @@ csharp_preferred_modifier_order = public, private, protected, internal, new, abs
 csharp_style_var_elsewhere = true:suggestion
 csharp_style_var_for_built_in_types = false:suggestion
 csharp_style_var_when_type_is_apparent = true:suggestion
-csharp_style_namespace_declarations = block_scoped:suggestion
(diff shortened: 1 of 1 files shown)


3fbaea00:
f82ca42 related: 
now instead of:
```cs
namespace AsmResolver;
class MyClass { }
```

visual studio generates:
```cs
namespace AsmResolver
{
    class MyClass { }
}
```

.editorconfig
```diff
@@ -13,6 +13,7 @@ csharp_preferred_modifier_order = public, private, protected, internal, new, abs
 csharp_style_var_elsewhere = true:suggestion
 csharp_style_var_for_built_in_types = false:suggestion
 csharp_style_var_when_type_is_apparent = true:suggestion
+csharp_style_namespace_declarations = block_scoped:suggestion
```
Washi1337, line 16: New files should prefer namespace declarations over blocks. We simply do not convert existing namespace blocks to declarations to not destroy git blame.

PR #779, opened 2026-08-15
Add Silverlight assembly resolution and runtime identification support

SychicBoy:
## Summary

This pull request adds first-class Microsoft Silverlight 4 and Silverlight 5 support to `AsmResolver.DotNet`.

Silverlight is introduced as a distinct target runtime with dedicated core library identities, installation discovery, and assembly resolution. `RuntimeContext` now selects a Silverlight-specific resolver when loading supported Silverlight assemblies.

The implementation follows the architecture discussed in #777 by introducing a dedicated `SilverlightAssemblyResolver` backed by a `SilverlightPathProvider`, while continuing to reuse `AssemblyResolverBase` for common resolution behavior.

## Implementation

This pull request:

* Adds Silverlight runtime identification to `DotNetRuntimeInfo`.
* Adds the Silverlight 4 and Silverlight 5 `mscorlib` identities to `KnownCorLibs`.
* Detects supported, unprofiled Silverlight targets from `TargetFrameworkAttribute`.
* Adds `SilverlightInstallation` and `SilverlightPathProvider`.
* Adds `MicrosoftSilverlightPathProvider` for discovering Silverlight reference assemblies, SDK client libraries, and installed runtime directories.
* Adds `SilverlightAssemblyResolver`, derived from `AssemblyResolverBase`.
* Integrates `SilverlightAssemblyResolver` with `RuntimeContext`.

The resolver probes Silverlight-specific locations in the following order:

1. Reference assemblies for the requested Silverlight version.
2. SDK client libraries for the requested Silverlight version.
3. A matching installed Silverlight runtime.
4. The standard locations handled by `AssemblyResolverBase`, including the originating module directory, working directory, and custom search directories.

Resolution remains target-version-specific; Silverlight 4 references are not implicitly rolled forward to Silverlight 5.

## Runtime Detection

Silverlight 4 and Silverlight 5 are detected from supported, unprofiled `TargetFrameworkAttribute` metadata.

Corlib identity alone is intentionally not used to classify an assembly as Silverlight. Some legacy retargetable and portable profiles share core library identities with Silverlight, which makes corlib-only detection ambiguous.

Profiled Silverlight targets, such as Windows Phone profiles, are therefore not routed through the desktop Silverlight resolver.

For assemblies where target-framework metadata is unavailable or has been removed, callers can explicitly construct a `RuntimeContext` for the desired Silverlight version.

## Testing

The test coverage is split between deterministic unit tests and optional installation-dependent integration tests.

The deterministic tests cover:

* Silverlight runtime parsing and corlib selection.
* Runtime probing and regression behavior.
* Silverlight `RuntimeContext` construction.
* Installation and path discovery using temporary directory layouts.
* Resolver probing order and fallback behavior.
* Originating-module, working-directory, and custom search-directory resolution.
* Configured `IFileService` usage.
* Synthetic Silverlight 4 and Silverlight 5 corlib resolution.

A Windows-only `SkippableTheory` additionally verifies resolution against real Silverlight 4 and Silverlight 5 reference assemblies when the corresponding SDK is installed. If the required SDK is unavailable, these integration cases are skipped rather than making Silverlight installation a requirement for the regular test suite.

## Scope

This pull request is intentionally limited to Silverlight 4 and Silverlight 5 desktop runtime detection and assembly resolution.

Support for older Silverlight runtime versions, platform-specific profiles such as Windows Phone, and deployment/hosting scenarios such as XAP processing and browser execution are not included in this change.

## Related

Discussion #777


src/AsmResolver.DotNet/ModuleDefinition.cs
```diff
@@ -354,6 +354,10 @@ public ModuleDefinition(Utf8String? name, AssemblyReference? corLib)
         public ModuleDefinition(Utf8String? name, DotNetRuntimeInfo targetRuntime)
             : this(name, targetRuntime.GetDefaultCorLib())
         {
+            // Silverlight corlib identities overlap with other legacy runtimes. Preserve the explicitly provided
+            // runtime instead of recomputing it from the corlib identity.
+            if (targetRuntime.IsSilverlight)
+                OriginalTargetRuntime = targetRuntime;
```
Washi1337: To be honest, we should probably just always set `OriginalTargetRuntime` to `targetRuntime`.

src/AsmResolver.DotNet/SilverlightAssemblyResolver.cs
```diff
@@ -0,0 +1,76 @@
...
+        if (_installation.SdkLibraryDirectory is { } sdkLibraryDirectory
+            && ProbeDirectory(assembly, sdkLibraryDirectory) is { } sdkLibraryPath)
+            return sdkLibraryPath;
+
+        if (_installation.RuntimeDirectory is { } runtimeDirectory
+            && ProbeDirectory(assembly, runtimeDirectory) is { } runtimePath)
+            return runtimePath;
+
+        return null;
+    }
```
Washi1337, line 61: To keep things consistent with other resolvers, we should prioritize runtime directories over reference/sdk. By default, AsmResolver always tries to mimic runtime behavior.

src/AsmResolver.DotNet/MicrosoftSilverlightPathProvider.cs
```diff
@@ -0,0 +1,221 @@
...
+            if (candidate.Version.Major != version.Major || candidate.Version.Minor != version.Minor)
+                continue;
+
+            installation = candidate;
+            return true;
+        }
+
+        installation = null;
+        return false;
+    }
```
Washi1337: I think this just finds the last matching version in the list of installations, but not necessarily also the newest compatible version (as happens in e.g., `DotNetFrameworkPathProvider`), because `_installations` is not necessarily sorted.

src/AsmResolver.DotNet/MicrosoftSilverlightPathProvider.cs
```diff
@@ -0,0 +1,221 @@
...
+
+                if (!File.Exists(Path.Combine(directory, "mscorlib.dll")))
+                    continue;
+
+                if (bestVersion is not null && candidateVersion <= bestVersion)
+                    continue;
+
+                bestVersion = candidateVersion;
+                bestDirectory = directory;
+            }
```
Washi1337, line 190: This only collects the latest framework in the directory of frameworks, not all of them.

I think this also highlights a more fundamental issue. We should probably not bunch up runtime and reference directories into one `SilverlightInstallation`. This would also make it more consistent with how `DotNetFxPathProvider` behaves, where there are functions querying for both runtime and reference installations.

src/AsmResolver.DotNet/MicrosoftSilverlightPathProvider.cs
```diff
@@ -0,0 +1,221 @@
...
+            yield return programFilesNative;
+
+        string programFiles = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles);
+        if (!string.IsNullOrEmpty(programFiles))
+            yield return programFiles;
+
+        string? programFilesEnvironment = Environment.GetEnvironmentVariable("ProgramFiles");
+        if (!string.IsNullOrEmpty(programFilesEnvironment))
+            yield return programFilesEnvironment;
+    }
```
Washi1337: This seems to always prefer 32-bit. Should we make this configurable, like in `DotNetFxPathProvider`?

src/AsmResolver.DotNet/TargetRuntimeProber.cs
```diff
@@ -187,17 +188,46 @@ private static bool TraverseTargetRuntimeAttribute(in MetadataStreamSelection st
...
-                && DotNetRuntimeInfo.TryParse(element, out var info)
-                && (bestMatch.Version is null || bestMatch.Name == info.Name && info.Version > bestMatch.Version))
+            if (Utf8String.IsNullOrEmpty(element) || !DotNetRuntimeInfo.TryParse(element, out var info))
+                continue;
+
+            // Silverlight shares corlib identities with other legacy target frameworks. Only use explicit, unprofiled
+            // Silverlight target framework metadata to classify an image as Silverlight.
+            bool isSupportedSilverlight = info.IsSilverlight
+                && IsSupportedSilverlightVersion(info.Version)
+                && !HasNonEmptyProfile(element);
```
Washi1337: I am not sure what a profiled/unprofiled Silverlight target means and/or why it affects the probing? Do we have a reference for this somewhere or was this determined experimentally?

Washi1337 (changes requested):
Thanks a lot for putting in the effort. There are a few things that I think still need addressing.

I am not sure we should put both reference and runtime assembly directories in one `SilverlightInstallation`. Considering them separate would make it more consistent with `DotNetFxPathProvider`. Furthermore, AsmResolver should always prefer runtime assemblies, since it tries to mimic what actually ends up happening (rather than what a compiler may think will happen). As such, I think `SilverlightAssemblyResolver` should therefore do the same thing as well.

We also need an addition to the `TargetRuntimeProberTest` that detects the use of silverlight in actual _binaries_. Ideally, we would have an actual silverlight binary that we can use as input.

Few other things here and there, see comments

SychicBoy:
Thanks for the detailed feedback. I have addressed the requested changes:

- Split Silverlight runtime and reference installations into separate `SilverlightInstallation` instances and separate path-provider lookups.
- Updated assembly resolution to prefer an installed runtime and use reference and SDK assemblies only when no compatible runtime is available.
- Added architecture-aware x86/x64 discovery, newest-compatible runtime selection, exact reference-version matching, and architecture fallback tests.
- Preserved explicitly supplied target runtimes in `ModuleDefinition`.
- Clarified that profiled Silverlight targets such as Windows Phone represent distinct platforms and should not be classified as desktop Silverlight.
- Added a real Silverlight 5 binary compiled with Visual Studio MSBuild against the Silverlight 5 targeting pack and covered it in `TargetRuntimeProberTest`.

src/AsmResolver.DotNet/TargetRuntimeProber.cs
```diff
@@ -187,17 +188,50 @@ private static bool TraverseTargetRuntimeAttribute(in MetadataStreamSelection st
...
+            // Silverlight shares corlib identities with other legacy target frameworks. Target framework profiles
+            // also identify distinct platforms, such as `Silverlight,Version=v4.0,Profile=WindowsPhone`, rather than
+            // the full desktop framework:
+            // https://learn.microsoft.com/visualstudio/extensibility/creating-a-software-development-kit
+            // Only use explicit, unprofiled Silverlight target framework metadata to classify an image as desktop
+            // Silverlight.
+            bool isSupportedSilverlight = info.IsSilverlight
+                && IsSupportedSilverlightVersion(info.Version)
+                && !HasNonEmptyProfile(element);
+
```
Washi1337: I am still not sure how this changes things. Does an explicit profile also change the installation directory used for runtime assembly lookup? As far as I know, the referenced link does not really explain it

src/AsmResolver.DotNet/TargetRuntimeProber.cs
```diff
@@ -227,6 +261,30 @@ public static DotNetRuntimeInfo ExtractDotNetRuntimeInfo(IResolutionScope corLib
...
+            string key = component.Substring(0, separatorIndex).Trim();
+
+            if (string.Equals(key, "Profile", StringComparison.OrdinalIgnoreCase)
+                && component.Substring(separatorIndex + 1).Trim().Length > 0)
+                return true;
+        }
+
+        return false;
+    }
```
Washi1337: Can we move `Profile` parsing into `DotNetRuntimeInfo` perhaps? Maybe we should add a `string? DotNetRuntimeInfo::Profile`. It seems out of place to put it here.

src/AsmResolver.DotNet/MicrosoftSilverlightPathProvider.cs
```diff
@@ -0,0 +1,266 @@
...
+                yield return programFilesEnvironment;
+        }
+    }
+
+    private static IEnumerable<string> GetDefaultProgramFiles64BitDirectories()
+    {
+        string? programFilesNative = Environment.GetEnvironmentVariable("ProgramW6432");
+        if (!string.IsNullOrEmpty(programFilesNative))
+            yield return programFilesNative;
+    }
```
Washi1337, line 248: This doesn't seem correct.`SpecialFolder.ProgramFiles` returns the non-x86 program files on non-x86 hosts.

Also, GetFolderPath already obtains the folder path through environment variables. As such, checking the environment variable after it is redundant.

src/AsmResolver.DotNet/MicrosoftSilverlightPathProvider.cs
```diff
@@ -0,0 +1,266 @@
...
+        IEnumerable<string> programFiles32BitDirectories,
+        IEnumerable<string> programFiles64BitDirectories)
+    {
+        if (programFiles32BitDirectories is null)
+            throw new ArgumentNullException(nameof(programFiles32BitDirectories));
+        if (programFiles64BitDirectories is null)
+            throw new ArgumentNullException(nameof(programFiles64BitDirectories));
+
+        string[] roots32 = GetDistinctDirectories(programFiles32BitDirectories);
+        string[] roots64 = GetDistinctDirectories(programFiles64BitDirectories);
```
Washi1337: Is there a difference between this and just calling `directories.Distinct().ToArray()`?

Washi1337 (changes requested):
Thanks!

However, I am still not sure I understand the explicit check for the `Profile` field in the target framework string. The referenced link does not really explain it to me.  

Also, testing this in a VM with a Silverlight 5.0 installation, I noticed that opening/resolving `mscorlib.dll` from the silverlight runtimes does not set `OriginalTargetRuntime` to Silverlight but NetCoreApp. This probably requires a small change in `TargetRuntimeProber::ToDotNetRuntimeInfo`, as this binary does not have a `TargetRuntimeAttribute` added to its assembly definition and thus AsmResolver has to rely on the assembly descriptor itself.

Few minor things, but other than that, LGTM.

src/AsmResolver.DotNet/TargetRuntimeProber.cs
SychicBoy: The referenced Microsoft documentation does explain this case explicitly:

> If a specific target framework profile is specified, only that profile will be used by the Reference Manager for filtering purposes. For example, when "Silverlight, version=v4.0, profile=WindowsPhone" is specified, Reference Manager filters on only the Windows Phone profile; a project targeting the full Silverlight 4.0 Framework does not see the SDK in the Reference Manager.

https://learn.microsoft.com/en-us/visualstudio/extensibility/creating-a-software-development-kit

So a profile such as `WindowsPhone` is treated as a different Silverlight target rather than the normal desktop Silverlight framework. Since the Silverlight resolver in this PR only probes the standard desktop Silverlight runtime and reference directories, profiled targets should not be classified as desktop Silverlight.

Some related references on how target framework profiles are used when locating reference assemblies:

https://learn.microsoft.com/en-us/visualstudio/msbuild/getreferenceassemblypaths-task

https://learn.microsoft.com/dotnet/api/microsoft.build.utilities.toollocationhelper.getpathtoreferenceassemblies

Washi1337: Ah OK, I must have missed this.

To be pedantic, this only suggests it affects reference/SDK assembly selection and not necessarily the selection at runtime. However, for the sake of this PR, let's assume for now these are indeed different _runtimes_ as well.

SychicBoy:
I’ve addressed the remaining review points.

- Target framework profiles are now parsed and stored by `DotNetRuntimeInfo`, instead of being parsed inside `TargetRuntimeProber`. 
- The explicit profile-based Silverlight rejection has been removed.
- Silverlight installation discovery was corrected for the 32-bit and 64-bit Program Files locations, and duplicate search paths are removed.
- Runtime probing now recognizes Silverlight assemblies from their mscorlib reference version when no `TargetFrameworkAttribute` is present. mscorlib version 2.0.5.0 is identified as Silverlight 4, and version 5.0.5.0 as Silverlight 5.
- Tests were added and updated to cover these cases.

Washi1337 (approved):

Washi1337:
Thanks!

PR #788, opened 2026-09-19
Support enum-backed FieldRVA data without class layouts

### LINKED ISSUE #786: FieldRvaDataReader cannot read enum RVA fields without ClassLayout
### AsmResolver Version

6.0.1

### .NET Version

.NET 10.0.12

### Operating System

Windows

### Describe the Bug

`FieldRvaDataReader` cannot read an RVA field when its type is an enum without a `ClassLayout` row.

An enum’s size can be determined from its underlying instance field, such as `System.Int32` or `System.Int64`, so it does not need a separate `ClassLayout` row. `FieldRvaDataReader` currently relies on `ClassLayout` to determine the size of custom value types stored in RVA fields and therefore rejects valid enum-backed RVA fields that the CLR accepts.

I encountered this in a C++/CLI executable built with MSVC. It contains several `System.Int32`-backed enum RVA fields without `ClassLayout` rows.

### How To Reproduce

The attached `FieldRvaEnum.zip` contains the IL source and the assembled executable.

Running `FieldRvaEnum.exe` prints `42`, confirming that the CLR accepts the enum-backed RVA field.

The executable can also be rebuilt from the included source using ILASM:

```powershell
ilasm FieldRvaEnum.il /exe /output=FieldRvaEnum.exe
```

Create a console project referencing `AsmResolver.DotNet` 6.0.1, then try to read the field’s RVA data and rebuild the module:

```csharp
using System.Linq;
using AsmResolver;
using AsmResolver.DotNet;
using AsmResolver.DotNet.Serialized;

var parameters = new ModuleReaderParameters(ThrowErrorListener.Instance);
var module = ModuleDefinition.FromFile("FieldRvaEnum.exe", parameters);

var field = module.TopLevelTypes
    .Single(type => type.Name == "Program")
    .Fields
    .Single(field => field.Name == "Data");

var data = field.FieldRva!.WriteIntoArray();

module.Write("FieldRvaEnum.rebuilt.exe");
```

Accessing `FieldRva` throws before the module can be written.

[FieldRvaEnum.zip](https://github.com/user-attachments/files/32418801/FieldRvaEnum.zip)

### Expected Behavior

When an RVA field uses an enum type without a `ClassLayout` row, AsmResolver should determine its size from the enum’s underlying type.

The RVA data should be available through `FieldDefinition.FieldRva` and preserved when the module is rebuilt.

In the attached reproduction, the enum is backed by `System.Int32`, so the RVA data consists of four bytes containing the value `42`. The rebuilt executable should still print `42`.

### Actual Behavior

AsmResolver requires a `ClassLayout` row instead of using the enum’s underlying type. Accessing `FieldRva` fails with:

`System.BadImageFormatException: Field type 02000002 does not have a class layout attached to it.`

As a result, the RVA data cannot be read and the module cannot be rebuilt through this code path.

SychicBoy:
## Summary

Allow `FieldRvaDataReader` to read enum-backed RVA fields without a `ClassLayout` row.

## Current Behavior

Reading these fields fails because their size cannot be determined without `ClassLayout`.

## Updated Behavior

Enum sizes are determined from their underlying instance fields.

## Related Issues

Fixes #786

## Checklist

- [x] Followed code guidelines in [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] Added unit tests (when possible)

## AI Usage Disclosure

- [ ] AI was used in writing the code of this PR
- [ ] AI was used in writing the text of this PR
- [x] I am not a meat proxy


src/AsmResolver.PE/DotNet/Metadata/FieldRvaDataReader.cs
```diff
@@ -182,13 +190,91 @@ private static int GetCustomTypeSize(IErrorListener listener, TablesStream table
...
+                return listener.BadImageAndReturn<int>($"Enum type {typeToken} has an invalid field list.");
+            }
+
+            // An enum's size comes from its single instance field, even without a ClassLayout row.
+            FieldDefinitionRow? underlyingField = null;
+            for (uint rid = fieldRange.StartRid; rid < fieldRange.EndRid; rid++)
+            {
+                uint fieldRid = rid;
+                if (fieldRange.IsRedirected && !fieldRange.RedirectionTable.TryGetCell(rid, 0, out fieldRid))
+                    return listener.BadImageAndReturn<int>($"Enum type {typeToken} has an invalid field list.");
```
Washi1337: MetadataRange supports iterating over all RIDs through `foreach`. This should already redirect tokens as they are read from any potential `FieldPtr` table.

```csharp
foreach (var token in fieldRange)
{
   uint fieldRid = token.Rid;
   ...
}
```

src/AsmResolver.PE/DotNet/Metadata/FieldRvaDataReader.cs
```diff
@@ -182,13 +190,91 @@ private static int GetCustomTypeSize(IErrorListener listener, TablesStream table
...
+
+                if (fieldRid > 0x00FFFFFF || !fieldTable.TryGetByRid(fieldRid, out var field))
```
Washi1337: `fieldRid > 0x00FFFFFF` is already captured by `fieldTable.TryGetByRid`

Washi1337 (changes requested):
Thanks for looking into it. Few comments but other than that LGTM

src/AsmResolver.PE/DotNet/Metadata/FieldRvaDataReader.cs
```diff
@@ -182,13 +190,87 @@ private static int GetCustomTypeSize(IErrorListener listener, TablesStream table
...
+
+            // Field list bounds refer to FieldPtr rows when a redirection table is present.
+            uint maximumRid = (uint) (fieldRange.RedirectionTable?.Count ?? fieldTable.Count) + 1;
+            if (fieldRange.StartRid == 0
+                || fieldRange.StartRid > maximumRid
+                || fieldRange.EndRid < fieldRange.StartRid
+                || fieldRange.EndRid > maximumRid)
+            {
+                return listener.BadImageAndReturn<int>($"Enum type {typeToken} has an invalid field list.");
+            }
```
Washi1337: I don't think these checks are needed if we're using `MetadataRange`. The built-in enumerator and  `table.TryGetByRid` already do bounds checks.

SychicBoy: Sorry, I missed that one. It’s removed now.

Washi1337:
Thanks!

